PR #31, opened 2012-11-01
ditch hasOwnProperty, use Object.create instead to get a clean object

### LINKED ISSUE #30: addMapping with the name __proto__ doesn't work
Looks like `__proto__` can't be added, because https://github.com/mishoo/UglifyJS2/issues/30#issuecomment-9973228

mishoo:
close #30

### DIFF
--- a/lib/source-map/array-set.js
+++ b/lib/source-map/array-set.js
@@ -17,7 +17,7 @@ define(function (require, exports, module) {
    */
   function ArraySet() {
     this._array = [];
-    this._set = Object.create(null);
+    this._set = {};
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ashaffer:
Hey @mishoo and @fitzgen, I just ran the benchmarks using Object.create(null) instead of the string prefixing, and it basically cuts the 'serializing source maps' benchmarks in half.

Can we re-evaluate using Object.create(null) since that requirement seems to be 4 years old now? It'd be a huge boost to performance, and I can't imagine that supporting all ES3 browsers is still a critical requirement for you guys...is it?

fitzgen:
Works for me.

michaelficarra:
Supporting ES3 environments is no longer important to me.

PR #150, opened 2015-01-06
Replace `_mappings` array with data structure

### LINKED ISSUE #149: Optimize generator with respect to sorting mappings
This is along the same vein as https://github.com/mozilla/source-map/issues/148, but less drastic and specialized. 

See that there have already been thoughts in this direction here: https://github.com/mozilla/source-map/blob/master/lib/source-map/source-map-generator.js#L297-L299

1. Small gains from keeping mappings always-sorted as an invariant
2. Especially cheap insertion of mappings when added in-order

Implementation wise this could be implemented by using a custom data type for `_mappings`. 

Some sort of heap, or even a plain array with sorted insertion and using `push` for mappings beyond maximum seen (line, column).

crisptrutski:
Closes #149

Optimized for best case (called with sorted mappings), with comparable performance otherwise.

### DIFF
--- a/lib/source-map/mapping-list.js
+++ b/lib/source-map/mapping-list.js
@@ -1,0 +1,80 @@
+/* -*- Mode: js; js-indent-level: 2; -*- */
+/*
+ * Copyright 2014 Mozilla Foundation and contributors
+ * Licensed under the New BSD license. See LICENSE or:
+ * http://opensource.org/licenses/BSD-3-Clause
+ */
+if (typeof define !== 'function') {
+    var define = require('amdefine')(module, require);
+}
+define(function (require, exports, module) {
+
+  var util = require('./util');
+
+  /**
+   * Determine whether mappingB is after mappingA with respect to generated 
+   * position.
+   */
+  function generatedPositionAfter(mappingA, mappingB) {
+    // Optimized for most common case
+    var lineA = mappingA.generatedLine;
+    var lineB = mappingB.generatedLine;
+    var columnA = mappingA.generatedColumn;
+    var columnB = mappingB.generatedColumn;
+    return lineB > lineA || lineB == lineA && columnB >= columnA ||
+           util.compareByGeneratedPositions(mappingA, mappingB) <= 0;
+  }
+
+  /**
+   * A data structure to provide a sorted view of accumulated mappings in a
+   * performance conscious manner. It trades a neglibable overhead in general
+   * case for a large speedup in case of mappings being added in order.
+   */
+  function MappingList() {
+    this._array = [];
+    this._sorted = true;
+    // Serves as infimum
+    this._last = {generatedLine: -1, generatedColumn: 0};
+  }
+
+  /**
+   * Iterate through internal items. NOTE: order is NOT guaranteed.
+   */
+  MappingList.prototype.unsortedForEach = function MappingList_forEach() {
+    Array.prototype.forEach.apply(this._array, arguments);
+  };
+
+  /**
+   * Add the given source mapping.
+   *
+   * @param Object aMapping
+   */
+  MappingList.prototype.add = function MappingList_add(aMapping) {
+    var mapping;
+    if (generatedPositionAfter(this._last, aMapping)) {
+      this._last = aMapping;
+      this._array.push(aMapping);
+    } else {
+      this._sorted = false;
+      this._array.push(aMapping);
+    }
+  };
+
+  /**
+   * Returns the flat array representation of this structure.
+   * 
+   * WARNING: Returns internal data without copying, for performance
+   */
+  MappingList.prototype.toArray = function MappingList_toArray() {
+    if (this._sorted) {
+      return this._array;
+    } else {
+      // Sort runs in place
+      this._sorted = true;
+      return this._array.sort(util.compareByGeneratedPositions);
+    }
+  };
+
+  exports.MappingList = MappingList;
+
+});
--- a/lib/source-map/source-map-consumer.js
+++ b/lib/source-map/source-map-consumer.js
@@ -102,9 +102,9 @@ define(function (require, exports, module) {
                                                               smc.sourceRoot);
       smc.file = aSourceMap._file;
 
-      smc.__generatedMappings = aSourceMap._mappings.slice()
+      smc.__generatedMappings = aSourceMap._mappings.toArray()
         .sort(util.compareByGeneratedPositions);
-      smc.__originalMappings = aSourceMap._mappings.slice()
+      smc.__originalMappings = aSourceMap._mappings.toArray()
         .sort(util.compareByOriginalPositions);
 
       return smc;
--- a/lib/source-map/source-map-generator.js
+++ b/lib/source-map/source-map-generator.js
@@ -12,6 +12,7 @@ define(function (require, exports, module) {
   var base64VLQ = require('./base64-vlq');
   var util = require('./util');
   var ArraySet = require('./array-set').ArraySet;
+  var MappingList = require('./mapping-list').MappingList;
 
   /**
    * An instance of the SourceMapGenerator represents a source map which is
@@ -29,7 +30,7 @@ define(function (require, exports, module) {
     this._sourceRoot = util.getArg(aArgs, 'sourceRoot', null);
     this._sources = new ArraySet();
     this._names = new ArraySet();
-    this._mappings = [];
+    this._mappings = new MappingList();
     this._sourcesContents = null;
   }
 
@@ -109,7 +110,7 @@ define(function (require, exports, module) {
         this._names.add(name);
       }
 
-      this._mappings.push({
+      this._mappings.add({
         generatedLine: generated.line,
         generatedColumn: generated.column,
         originalLine: original != null && original.line,
@@ -186,7 +187,7 @@ define(function (require, exports, module) {
       var newNames = new ArraySet();
 
       // Find mappings for the "sourceFile"
-      this._mappings.forEach(function (mapping) {
+      this._mappings.unsortedForEach(function (mapping) {
         if (mapping.source === sourceFile && mapping.originalLine != null) {
           // Check if it can be mapped by the source map, then update the mapping.
           var original = aSourceMapConsumer.originalPositionFor({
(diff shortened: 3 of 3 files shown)


lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+}
+define(function (require, exports, module) {
+
+  var util = require('./util');
+
+  /**
+   * Determine whether mappingB is after mappingA with respect to generated position.
+   */
+  function generatedPositionAfter(mappingA, mappingB) {
+    // Optimized for most common case
```
crisptrutski: To justify this unrolling, it provided a slightly over 2x speedup in my testing. Roughly corresponds to the `onlyCompareGenerated` flag in the util function.

lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+    // Serves as infimum
+    this._last = {generatedLine: -1, generatedColumn: 0};
+  }
+
+  /**
+   * Static method for creating MappingList instances from an existing array.
+   */
+  MappingList.fromArray = function MappingList_fromArray(aArray) {
+    var list = new MappingList();
+    last._array = aArray.slice().sort(util.compareByGeneratedPositions);
```
lydell: l<strong>a</strong>st -> l<strong>i</strong>st ?

crisptrutski: Good catch - looks like this code is untested too. Perhaps redact this unused method even?

crisptrutski:
Let me know if I should squash the fixup commits

crisptrutski:
Here are the benchmarks I've run so far (couldn't get markdown table to render in comment for some reason, so excuse poor legibility)

### 1. Data structure performance

These all use 10,000 mappings.

#### Push to Array only (ie. unsorted at the end)

```
  11,440 ops/sec ±0.49% (83 runs sampled)
```

#### Push to Array then use .sort

```
  sorted input:   198 ops/sec ±1.08% (84 runs sampled)
  unsorted input: 191 ops/sec ±1.09% (87 runs sampled)
```

#### Push to MappingList then request sorted array

```
  sorted input:  3,105 ops/sec ±0.57% (92 runs sampled) <<-- oh yeah
  unsorted input 185 ops/sec ±0.93% (85 runs sampled)
```


### 2. Overall performance (use generator to produce JSON string)

#### Using Array for _mappings

**Sorted input**

```
10,000    @ 41.61 ops/sec ±0.95% (55 runs sampled)
100,000   @ 3.12 ops/sec ±2.14% (12 runs sampled)
1,000,000 @ 0.25 ops/sec ±2.27% (5 runs sampled)
```

**Unorted input**

```
10,000    @ 40.60 ops/sec ±1.38% (54 runs sampled)
100,000   @ 2.87 ops/sec ±2.32% (12 runs sampled)
1,000,000 @ 0.21 ops/sec ±5.91% (5 runs sampled)
```


#### Using MappingList for _mappings

**Sorted input**

```
10,000    @ 48.52 ops/sec ±0.98% (63 runs sampled)
100,000   @ 3.68 ops/sec ±2.18% (14 runs sampled)
1,000,000 @ 0.30 ops/sec ±3.89% (5 runs sampled)
```

**Unorted input**

```
10,000    @ 39.27 ops/sec ±1.50% (52 runs sampled)
100,000   @ 2.83 ops/sec ±1.97% (12 runs sampled)
1,000,000 @ 0.21 ops/sec ±5.26% (5 runs sampled)
```


#### Using MappingList for _mappings, and skipValidation on

**Sorted input**

```
10,000    @ 64.56 ops/sec ±1.08% (67 runs sampled)
100,000   @ 4.50 ops/sec ±1.69% (16 runs sampled)
1,000,000 @ 0.34 ops/sec ±13.41% (5 runs sampled)
```

**Unorted input**

```
10,000    @ 50.20 ops/sec ±0.82% (65 runs sampled)
100,000   @ 3.20 ops/sec ±3.46% (12 runs sampled)
1,000,000 @ 0.23 ops/sec ±6.82% (5 runs sampled)
```

fitzgen:
Great results, thank you for the detailed benchmarking!

Looking at the code now...

lib/source-map/mapping-list.js
fitzgen: > To justify this unrolling, it provided a slightly over 2x speedup in my testing. Roughly corresponds to the onlyCompareGenerated flag in the util function.

Can you add that as a comment then, so it is clear to future readers?

lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+if (typeof define !== 'function') {
+    var define = require('amdefine')(module, require);
+}
+define(function (require, exports, module) {
+
+  var util = require('./util');
+
+  /**
+   * Determine whether mappingB is after mappingA with respect to generated position.
```
fitzgen: Nitpick: please wrap comments at column 80

lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+      this._array.push(aMapping);
+    } else {
+      this._unordered = true;
+      this._array.push(aMapping);
+    }
+  };
+
+  /**
+   * Returns the flat array representation of this structure. Note that it also returns a shallow
+   * copy of the data to avoid tampering with internal state.
```
fitzgen: Nitpick: wrap @ 80

lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+    return list;
+  };
+
+
+  /**
+   * Iterate through internal items.
+   */
+  MappingList.prototype.forEach = function MappingList_forEach() {
+    return Array.prototype.forEach.apply(this._array, arguments);
+  };
```
fitzgen: This method is unused, right?

I think we should remove it -- it doesn't guarantee sorted order and I think that is a potential foot gun in the future. If we don't need it, let's remove it.

lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * A data structure to provide a sorted view of accumulated mappings in a
+   * performance conscious manner. It trades a neglibable overhead in general
+   * case for a large speedup in case of mappings being added in order.
+   */
+  function MappingList() {
+    this._array = [];
+    this._unordered = false;
```
fitzgen: Nitpick: instead of the double negative, let's be more positive:

    this._ordered = true;

I think this reads better.

lib/source-map/mapping-list.js
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Returns the flat array representation of this structure. Note that it also returns a shallow
+   * copy of the data to avoid tampering with internal state.
+   */
+  MappingList.prototype.toArray = function MappingList_toArray() {
+    if (this._unordered) {
+      // Sort runs in place
+      this._unordered = false;
+      return this._array.sort(util.compareByGeneratedPositions);
```
fitzgen: Still need to `slice()` like in the else branch, the return value of `sort()` is the same array instance:

    > a = [3,45,7]
    Array [ 3, 45, 7 ]
    > b = a.sort()
    Array [ 3, 45, 7 ]
    > a == b
    true
    > a.pop()
    7
    > b
    Array [ 3, 45 ]

fitzgen:
This looks great! Thank you for taking the time to match the existing style, etc. It is appreciated.

Will merge the PR once you rebase+squash and fix up the small comments I had.

As a follow up, it seems that once we call `toArray` on a `MappingList`, we never use the `MappingList` again. Taking advantage of this observation, we could add a `release` method to `MappingList` which invalidates the `MappingList` instance and returns its internal array. This skips the array allocation and the O(n) copying of items into the new array that happens when we return `this._array.slice()`. We would need to add runtime assertions that we don't use that `MappingList` instance again afterwards. Rather than adding an `if` at the start of every method, which would be a slow down when building up the `MappingList`, we could monkey patch each method with a new function that just throws an assertion failure error.

lib/source-map/mapping-list.js
crisptrutski: So, it does get used here https://github.com/crisptrutski/source-map/blob/optimistic-add-mapping/lib/source-map/source-map-generator.js#L190. 

Used to make local mutations to the elements, so no cares about order. Could skip return value in wrapper to obliviate one self-harm vector, but yeah it'd be confusing with functions with non-scoped side effects.

To be honest I don't see much scope for reuse of this class though - passing in comparator as function was too much overhead so now it's really tightly coupled and just an implementation detail of the generator.

How about an ignoble `this._mappings._array.forEach` there and killing the wrapper method? You could consider `_array` to be in "friend" scope :goat:

lib/source-map/mapping-list.js
michaelficarra: "sorted" is a more appropriate term than "ordered":

```js
this._sorted = true;
```

lib/source-map/mapping-list.js
crisptrutski: Wanted to sort in place for the amortizing behaviour, but yeah missing a `slice` after the sort, oops. You're probably right about finalizing the class after toArray though - in which case we then don't need the slice again (maybe I was prescient after all)

lib/source-map/mapping-list.js
fitzgen: How about we just rename it `unsortedForEach` or `noSortForEach` so that it doesn't imply any sorting order? Should add a blurb in the doc comment about how it doesn't ensure sorting.

crisptrutski:
`release` makes it sounds like the concern i memory - should I be removing the array reference also? Not  convinced this is necessary, think it made sense for the data to have the same lifespan as the generator. 

How about `finalize` or `seal`? It would leave the array in place, if necessary sort in-place and set `_sorted` to true, and just monkey patch `add`.

Sound about right?

crisptrutski:
Actually `release` and `finalize` both sounds like memory management, and `seal` sounds like inheritance. Going with `freeze` in first spin.

fitzgen:
The lifetimes are _not_ the same because as soon as you give out the internal array, the instance is invalidated. If we had something like rust's immutable borrows, then maybe not, but...

I like `release` because you are releasing ownership of the array. Maybe `releaseArray`?

crisptrutski:
Actually, think we should abort this last change - it's a breaking change (see `test ignore duplicate mappings.` test). I'm actually fine with skipping the `slice` WITHOUT breaking `add`, because the array never escapes the generator.

fitzgen:
`releaseArrayAndInvalidate`?

crisptrutski:
Well, repeated calls to serialize the generator would reuse the _mapping instance, so more work would be required to decouple their lifetimes.

fitzgen:
Ah good point. I guess we should just give up on the idea of enforcing safety here then and just trust `SourceMapGenerator` to do the right thing.

crisptrutski:
Think all points have been addressed - let me know if you're satisfied and I'll squash 'em all tomorrow

fitzgen:
Yup looks good, thanks! Let me know when you've squashed.

crisptrutski:
@fitzgen squasherized on latest master

fitzgen:
Thanks!

crisptrutski:
Thanks for the library and thanks for accepting my self-serving amendments :)

fitzgen:
No problem, 0.1.43 published.

Issue #148, opened 2014-12-22
Cheap and greedy generator for common usage pattern


crisptrutski:
Hey,

Busy looking at improving build performance, specifically heavy use is being made of Traceur and compile times increase dramatically when sourcemap generation is enabled.  

The usage pattern of `SourceMapGenerator` via Traceur is very predicable:
1. Mappings always are always of the "size 4" variety (no name)
2. Sources are all known in advance
3. Mappings are all added in-order

Seems possible one could encode mapping strings greedily, skipping edge cases, intermediate representation, deferred computation etc.

Also imagine this usage pattern to be very common in other compilers. (The assumption of "no names" could relax to "name list known in advance" with only marginal extra overhead)

Perhaps something like the following:

``` javascript
  class OptimizedSourceMapGenerator {
    // Note that `sources` and `names` are now required
    constructor(file, sourceFileRoot, sources, names, [sourcesContents])

    // Assume (or assert) called with (generatedLine, generatedColumn) strictly ascending
    // This is the only retained public method from regular implementation, beyond toString, toJSON
    addMapping(generatedLine, generatedColumn, originalLine, originalColumn, source, name)
  }
```

Not sure how much this would impact performance, but have a hunch that it could be large! 

Would appreciate feedback on whether:
1. There seems to be merit in this approach, or whether you'd anticipate little improvement
2. You have other suggestions
3. PR welcome here for this, or best left as a 3rd party extension.

fitzgen:
Another possible optimization is to only generate one mapping per line (if you aren't generating minified JS).

Most debuggers step on each line, and so multiple mappings for a single generated line are usually extraneous.

This sped up [pretty-fast](https://github.com/mozilla/pretty-fast)'s source map generation a ton.

crisptrutski:
Hey, Nic

Thanks for so much feedback and input!

1. I'm not keen on adding extra addMappingSorted API point. This means updating Traceur also, adding extra test cases here, and it looks like jhttps://github.com/mozilla/source-map/issues/149 can get effectively the same speedup (went with a solution involving deoptimizing once out-of-order mapping added, and this overhead adds about 3-4% overhead, from benchmarking up to 1M mappings added)

2. Yes my assumption was that GC was an issue, but my attempts at measuring it have been crude - just calculating the total time spent in garbage collection. 

[...]

PR #151, opened 2015-01-06
Add option to skip mapping validation in generator

crisptrutski:
A performance boosting option for use in trusted contexts.

Extra context: #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,10 @@ You may pass an object with the following properties:
 
 * `sourceRoot`: A root for all relative URLs in this source map.
 
+* `skipValidation`: Optional flag to disable validation of mappings as they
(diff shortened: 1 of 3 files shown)


fitzgen:
Can you add documentation to `README.md` describing this option, with a warning that using it is generally unsafe and should only be used if you really need to eek out the last bit of performance, and should probably never be passed in tests.

Will merge with that.

cc @mishoo, who wanted something like this some time ago.

crisptrutski:
Great point about turning this off in tests, I'd have flubbed that.

crisptrutski:
@fitzgen documented

fitzgen:
Thanks!

PR #152, opened 2015-01-06
Remove RegExes in SourceNode.toStringWithSourceMap

danielmendel:
Faster implementation that passes all tests.  I made some [jsperf.com benchmarks](http://jsperf.com/match-vs-replace-split) that measure the discrete changes.  The remaining use of `REGEX_NEWLINE` could be replaced with the same `replace("\r\n","\n").split('')` approach and be *supremely* faster in SpiderMonkey, but slightly slower in V8.  Given that it's likely most source map generation is happening in node.js processes it seems better to optimize for that case.

### DIFF
--- a/lib/source-map/source-node.js
+++ b/lib/source-map/source-node.js
@@ -16,8 +16,8 @@ define(function (require, exports, module) {
   // operating systems these days (capturing the result).
   var REGEX_NEWLINE = /(\r?\n)/;
 
-  // Matches a Windows-style newline, or any character.
-  var REGEX_CHARACTER = /\r\n|[\s\S]/g;
+  // Newline character code for charCodeAt() comparisons
+  var NEWLINE_CODE = 10;
 
   // Private symbol for identifying `SourceNode`s when multiple versions of
   // the source-map library are loaded. This MUST NOT CHANGE across
@@ -375,12 +375,12 @@ define(function (require, exports, module) {
         lastOriginalSource = null;
         sourceMappingActive = false;
       }
-      chunk.match(REGEX_CHARACTER).forEach(function (ch, idx, array) {
-        if (REGEX_NEWLINE.test(ch)) {
+      for (var idx = 0, length = chunk.length; idx < length; idx++) {
+        if (chunk.charCodeAt(idx) === NEWLINE_CODE) {
           generated.line++;
           generated.column = 0;
           // Mappings end at eol
-          if (idx + 1 === array.length) {
+          if (idx + 1 === length) {
             lastOriginalSource = null;
             sourceMappingActive = false;
           } else if (sourceMappingActive) {
@@ -398,9 +398,9 @@ define(function (require, exports, module) {
             });
           }
         } else {
-          generated.column += ch.length;
+          generated.column++;
         }
-      });
+      }
     });
     this.walkSourceContents(function (sourceFile, sourceContent) {
       map.setSourceContent(sourceFile, sourceContent);


sokra:
There is no testcase for this, but as replace(string, string) only replaces the first occurence this could fail:
``` js
new SourceMap(null, null, null, "a\r\nb\r\n").toStringWithSourceMap({ file: "" })
```

danielmendel:
@sokra Whoops! Can't believe I forgot about that.  I've updated the PR to use a faster regex to do the replace globally in 0c4c737.  See these [updated benchmarks](http://jsperf.com/newline-normalization-and-char-array).

fitzgen:
@danielmendel looks good, can you rebase+squash and add a test to cover the case that @sokra mentioned?

I think something like this should cover it:

    var res = new SourceNode(null, null, null, [
      "\r\n\r\n\r\n",
      new SourceNode("a.js", 1, 0, "hi()")
    ]).toStringWithSourceMap();
    // assert that the mapping for `hi()` is on line 4

I will merge the PR once you do that.

As a follow up, if you're interested: I suspect there are further wins to be had from using a straight up `while` loop with an incrementing index instead of splitting and then using `forEach`. That way we would only do a single pass over the string.

lib/source-map/source-node.js
```diff
@@ -375,8 +373,8 @@ define(function (require, exports, module) {
         lastOriginalSource = null;
         sourceMappingActive = false;
       }
-      chunk.match(REGEX_CHARACTER).forEach(function (ch, idx, array) {
-        if (REGEX_NEWLINE.test(ch)) {
+      chunk.replace(REGEX_WIN_NEWLINE, "\n").split('').forEach(function (ch, idx, array) {
+        if (ch === "\r\n" || ch === "\n") {
```
lydell: This check should just be `if (ch === "\n")` now. Also, further down you can replace `+= ch.length` with `+= 1` (or `++`.)

fitzgen: Good catch, thanks!

lydell:
> I suspect there are further wins to be had from using a straight up while loop with an incrementing index instead of splitting and then using forEach.

+1. Possibly something like:

```js
var length = chunk.length;
var ch;
var lastCh;
for (var index = 0; index < length; index++) {
  var ch = chunk[index];
  if (ch === "\r" || (ch === "\n" && lastCh !== "\r")) {
  ...
  if (idx + 1 === length) {
  ...
  lastCh = ch;
```

danielmendel:
@lydell @fitzgen agreed - e10ea94 is a squashed commit that moves entirely to a for loop.  Tried to write a test for multiple windows-style newlines, but it wouldn't fail under the `replace` case above.  Turns out the incomplete replacer still generates the correct number of lines because `generatedLines` is only incremented on `"\n"` in every discussed implemenation.  The failure there would actually cause the `generatedColumn` count to be off, and then only for characters appearing between `"\r"` and `"\n"`.

lydell:
Nice! (A lot nicer than my quick sample above.)

> The failure there would actually cause the generatedColumn count to be off, and then only for characters appearing between "\r" and "\n".

We can actually assume that there will _never_ be any characters between "\r" and "\n", because from the beginning I added support for bare "\r" newlines, but then we decided to remove that and only support "\r\n" and "\n" newlines, since nothing else is commonly used anymore. So perhaps you can simplify the code and only do `if (ch === "\n")`?

lib/source-map/source-node.js
```diff
@@ -375,12 +372,16 @@ define(function (require, exports, module) {
         lastOriginalSource = null;
         sourceMappingActive = false;
       }
-      chunk.match(REGEX_CHARACTER).forEach(function (ch, idx, array) {
-        if (REGEX_NEWLINE.test(ch)) {
+      for (var idx = 0, length = chunk.length; idx < length; idx++) {
+        ch = chunk[idx];
+        if (ch === "\r" && chunk[idx+1] === "\n") {
+            continue;
+        }
```
danielmendel: @lydell sure thing - you mean straight up remove lines `377-379`?  Seems reasonable.

lydell: Oops, sorry—yes, that’s what I meant.

lydell:
And if we’re talking real performance optimization here: It _might_ be faster to use `if (chunk.charCodeAt(index) === NEWLINE)`, where `var NEWLINE = 10`.

danielmendel:
@lydell :+1: - I like it! This is a verifiably hot path for us right now so I'll take every extra cycle I can get.  Updated in 5485d40.

danielmendel:
Also here are some benchmarks on `charCodeAt` for the curious: http://jsperf.com/string-compare-vs-char-code-at

fitzgen:
Awesome, thank you! This looks good to me, @danielmendel  was there anything else you planned on changing before I merge?

danielmendel:
@fitzgen my pleasure - nothing more to add unless someone else chimes in, merge away.

PR #154, opened 2015-01-27
Fuzzy search should find the closest element that is greater than or equal to

ejpbruel:


lib/source-map/source-map-consumer.js
```diff
@@ -446,14 +446,12 @@ define(function (require, exports, module) {
...
-      // returns the index of the closest mapping less than the needle. By
-      // setting needle.originalColumn to Infinity, we thus find the last
-      // mapping for the given line, provided such a mapping exists.
+      var originalLine = util.getArg(aArgs, 'line');
+
       var needle = {
         source: util.getArg(aArgs, 'source'),
-        originalLine: util.getArg(aArgs, 'line'),
-        originalColumn: Infinity
+        originalLine: originalLine + 1,
```
fitzgen: Why +1?

fitzgen:
Looks good overall, but I'd like to see your response to my comment / updated code, and this needs to be rebased on master.

ejpbruel:
I've addressed your comments and rebased on top of master.

Note that indexed source maps actually rely on binary search returning the closest element that is less than the one we are searching for. We still want to be able to find the closest element that is greater than the one we are searching for finding specific mappings within a single map, however.

Since we need different modes of fuzzy search, I added a bias parameter to the binary search algorithm that determines whether the closest element greater than or less than the element we are searching for is returned when no exact element can be found.

fitzgen:
Did you forget about the `TODO`s in the tests?

Also, can we define a poor man's enum for the binary search fuzzy direction bias like we do with `SourceMapConsumer.{GENERATED,ORIGINAL}_ORDER`? Calling functions with booleans isn't super readable :-P

lib/source-map/source-map-consumer.js
```diff
@@ -199,10 +198,14 @@ define(function (require, exports, module) {
                                     this._originalMappings,
                                     "originalLine",
                                     "originalColumn",
-                                    util.compareByOriginalPositions);
+                                    util.compareByOriginalPositions,
+                                    true);
```
fitzgen: I don't think this `true` needs to be passed here. Left over from earlier iteration? Confirm and I can merge and remove it.

fitzgen:
ping @ejpbruel

ejpbruel:
Confirmed.

Is there some way I can make Github notify me when somebody comments on one of my pull requests? I  didn't see your last comment until I checked back here.

past:
Settings -> Notification Center -> Participating

sokra:
Maybe a bit late, but I think there is something wrong with this commits...

The behavior of `originalPositionFor` changed:

``` js
    var map = new SourceMapGenerator({
      file: 'foo.js'
    });
    map.addMapping({
      original: { line: 1, column: 1 },
      generated: { line: 2, column: 2 },
      source: 'bar.js'
    });
    map.addMapping({
      generated: { line: 2, column: 10 }
    });
    // 0 1 2 3 4 5 6 7 8 9 10 11
    //     [----bar.js----]
    // ^   ^     ^
    // a   b     c
    map = new SourceMapConsumer(map.toString());

    var pos;

    // a)
    pos = map.originalPositionFor({
      line: 2,
      column: 0,
    }); // => { source: 'bar.js', line: 1, column: 1, name: null }
    // but I would expect this to be
    // { source: null, line: null, column: null, name: null }

    // b)
    pos = map.originalPositionFor({
      line: 2,
      column: 2,
    }); // => { source: 'bar.js', line: 1, column: 1, name: null }
    // That's fine

    // c)
    pos = map.originalPositionFor({
      line: 2,
      column: 5,
    }); // => { source: null, line: null, column: null, name: null }
    // but I would expect this to be
    // { source: 'bar.js', line: 1, column: 1, name: null }
```

ejpbruel:
Sokra, that was the intention of this patch. It changes the fuzzy search algorithm so that if we can't find an exact match, we try to find the closest element that is *larger than* the one we are looking for (it used to be the closest element that is *smaller than*).

Taking c) as example, using the new algorithm, we are looking for the original position of line 2, column 5 in the generated source. There is no exact mapping for that position, so we find the closest one that is larger than the one we are looking for, which is the mapping for line 2, column 10 in the generated source. That mapping doesn't contain an original location however, which is why the result is the null mapping.

Having said that, I've since realized that we probably *don't* want to change the behavior of originalPositionFor (only for generatedPositionFor) so I'm about to file a pull request that will revert this change. Once that lands, your examples should work the way they did before.

sokra:
What's your use case? According to the SourceMap spec a mapping denote the starting.

ejpbruel:
The use case for this arised from a bug we are currently seeing in the Firefox developer tools. When you set a breakpoint on column 0 of a given line in a source mapped source, we expect that breakpoint to end up at the start of that line. However, if that line happens to be indented, the start of the line is not at column 0, and not all source maps provide a mapping for column 0 in that case.

When we don't have an exact mapping, the fuzzy search algorithm is used. Before this patch, this would give you the closest mapping that is smaller than the one we are looking for. Since we are already on column 0, this mapping is on the previous line, and we end up setting the breakpoint there.

What you'd expect when setting a breakpoint on column 0 of an indented line is that you find the closest mapping that is greater than the one we are looking for, which should be the first non-whitespace column of the indented line. There was no particular reason for the fuzzy search algorithm to return the closest mapping that is smaller than instead, hence this change.

Note that this change does not violate the spec, since all mappings still denote the start. The only thing that changed is the behavior of the source map library when looking for a mapping for which we do not have an exact match, which is not defined by the spec.

sokra:
> Before this patch, this would give you the closest mapping that is smaller than the one we are looking for.

Yeah, this is what I would expect.

The spec say: `The zero-based starting column of the line in the generated code that the segment represents`

> There was no particular reason for the fuzzy search algorithm to return the closest mapping that is smaller than instead, hence this change.

Of course there is. When the whole line is mapped (by a mapping on column 0), asking for orignal source of column 5 shouldn't return `null`.

Also ´SourceMapGenerator.applySourceMap` relies on this behavior and is broken now (unfortually there is no test case of this).

> The only thing that changed is the behavior of the source map library when looking for a mapping for which we do not have an exact match, which is not defined by the spec.

Yes, maybe the spec is not clear about this, but I assumed that mappings are active until a new mapping or a newline.

---

I would prefer to restore the original behavior and add a new function `fuzzyOriginalSourceFor` which tries `originalSourceFor` first and uses the next mapping when `null` is returned.

ejpbruel:
You are right. Mappings are active until a new mapping or a new line. When the whole line is mapped by a mapping that starts at column 0, searching for the mapping for column 5 should give you the one for column 0.

It's less clear what happens when an indented line is mapped by a mapping that starts at column 4, say. Currently, searching for the mapping for column 0 returns the last mapping for the previous line in that case. But as you pointed out, that mapping is no longer active after the new line.

I actually ran into this problem when trying to find the generated position of a breakpoint set on column 0 on an indented line in a generated source. The breakpoint ended up on the previous line, because that's where the closest mapping that is smaller than the one I am looking for is. What I really want in my case is the closest mapping that is larger than the one I am looking for.

It looks like the correct behavior original/generatedSourceFor when using fuzzy search is dependent on the use case, so how about adding an additional parameter to these functions that determines the search bias, and which defaults to the old behavior?

sokra:
> Currently, searching for the mapping for column 0 returns the last mapping for the previous line in that case. But as you pointed out, that mapping is no longer active after the new line.

This seems incorrect. It should return `null`.

> It looks like the correct behavior original/generatedSourceFor when using fuzzy search is dependent on the use case, so how about adding an additional parameter to these functions that determines the search bias, and which defaults to the old behavior?

yep, sounds good.

sokra:
You better revert the changes to the tests in this PR.

sokra:
Here is a testcase for the broken applySourceMap behavior:

``` js
  exports['test applySourceMap with unexact match'] = function (assert, util) {
    var map1 = new SourceMapGenerator({
      file: 'bundled-source'
    });
    map1.addMapping({
      generated: { line: 1, column: 4 },
      original: { line: 1, column: 4 },
      source: 'transformed-source'
    });
    map1.addMapping({
      generated: { line: 2, column: 4 },
      original: { line: 2, column: 4 },
      source: 'transformed-source'
    });

    var map2 = new SourceMapGenerator({
      file: 'transformed-source'
    });
    map2.addMapping({
      generated: { line: 2, column: 0 },
      original: { line: 1, column: 0 },
      source: 'original-source'
    });

    var expectedMap = new SourceMapGenerator({
      file: 'bundled-source'
    });
    expectedMap.addMapping({
      generated: { line: 1, column: 4 },
      original: { line: 1, column: 4 },
      source: 'transformed-source'
    });
    expectedMap.addMapping({
      generated: { line: 2, column: 4 },
      original: { line: 1, column: 0 },
      source: 'original-source'
    });

    map1.applySourceMap(new SourceMapConsumer(map2.toJSON()));

    util.assertEqualMaps(assert, map1.toJSON(), expectedMap.toJSON());
  };
```

fitzgen:
The issue is the debugger is operating with the wrong primitives: its searching for the One True Mapping, when it needs to just get all mappings for the target line. That is the end-goal, but Eddy is making incremental progress because breakpoint setting code is hairy and full of edge cases.

That said, the fact that the debugger is incrementally moving to a line-based rather than individual-mapping-based approach shouldn't break the individual-mapping-based translation. This is my bad: I should have caught this in review / before these patches even got written.

> It looks like the correct behavior original/generatedSourceFor when using fuzzy search is dependent on the use case, so how about adding an additional parameter to these functions that determines the search bias, and which defaults to the old behavior?

Yes, let's do this. And yes, default to the old behavior.

Additionally, can you include the new test case @sokra provided, @ejpbruel ? Thank you very much.

PR #157, opened 2015-02-04
make toString explicitly call toJSON, fixes #156

### LINKED ISSUE #156: generator.toString() versus JSON.stringify(generator.toJSON())
The instructions in the readme for Generating a Source Map using the low level api state that a client should create a generator, call `addMapping` as needed, and then call `toString()` to get the map out follows:

```
console.log(map.toString());
// '{"version":3,"file":"source-mapped.js","sources":["foo.js"],"names":["christopher"],"mappings":";;;;;;;;;mCAgCEA"}'
```

However, the generator toString is defined as:
```
    function SourceMapGenerator_toString() {
      return JSON.stringify(this);
    };
```

Which does not generate the form in the comment below the console log, it generates something like:
```
{ _file: 'sourcemap.js.map',
  _sourceRoot: null,
  _sources: { _array: [ 'fixture.js' ], _set: { '$fixture.js': 0 } },
  _names: { _array: [ 'map', 'v' ], _set: { '$map': 0, '$v': 1 } },
  _mappings: 
   [ { generatedLine: 1,
       generatedColumn: 0,
       originalLine: 1,
       originalColumn: 0,
       source: 'fixture.js',
       name: null },
     { generatedLine: 1,
       generatedColumn: 3,
       originalLine: 1,
       originalColumn: 3,
       source: 'fixture.js',
       name: 'map' }, ],
  _sourcesContents: { '$fixture.js': '[].map(v => v + 1)' } }
```

So, as far as I can tell, either the readme needs to change to instruct clients to call `JSON.stringify(map.toJSON())`, or the definition of `generator.toString` needs to change to return the correct source map format.

I am happy to submit either patch, but I wanted to raise this as an issue in case I'm missing something here, and/or to ask which approach is preferable. Personally, I think the definition of toString should change as that seems to be what was intended in the first place, but I only have a pony in this race.

brycecr:
If the client system has remapped JSON to something else,
`generator.toString()` may not call `generator.toJSON()`,
so this change makes the call to `toJSON()` explicit to
make toString more robust in this case.

### DIFF
--- a/lib/source-map/source-map-generator.js
+++ b/lib/source-map/source-map-generator.js
@@ -392,7 +392,7 @@ define(function (require, exports, module) {
    */
   SourceMapGenerator.prototype.toString =
     function SourceMapGenerator_toString() {
-      return JSON.stringify(this);
+      return JSON.stringify(this.toJSON());
     };
 
   exports.SourceMapGenerator = SourceMapGenerator;


fitzgen:
Cool, thanks!

PR #159, opened 2015-02-18
Improve the performance of parseMappings.

ejpbruel:
Based on our conversation yesterday, I wanted to see if I could come up with ways to improve the performance of parseMappings. I came up with two ideas:

1. We create a new string object for each iteration of the parse loop, using str.slice. Avoiding this leads to a 3% performance increase on my machine.

2. Huge source maps such as the scalajs-runtime.js.map show a lot of homogeneity between different segments (since we encode each offset relative to the previous one, this is what I would expect). Caching the parsed variable length fields for each segment leads to another 5% performance increase on my machine.

These are decent, but not huge wins. Moreover, there is some risk that 2. actually decreases performance in huge source maps where there is very little homogeneity between segments (although I expect such cases to be rare). I'll leave it up to you to decide whether it's worth adding these patches to the source-map library.

### DIFF
--- a/lib/source-map/base64-vlq.js
+++ b/lib/source-map/base64-vlq.js
@@ -117,18 +117,17 @@ define(function (require, exports, module) {
    * Decodes the next base 64 VLQ value from the given string and returns the
    * value and the rest of the string via the out parameter.
    */
-  exports.decode = function base64VLQ_decode(aStr, aOutParam) {
-    var i = 0;
+  exports.decode = function base64VLQ_decode(aStr, aIndex, aOutParam) {
     var strLen = aStr.length;
     var result = 0;
     var shift = 0;
     var continuation, digit;
 
     do {
-      if (i >= strLen) {
+      if (aIndex >= strLen) {
         throw new Error("Expected more digits in base 64 VLQ value.");
       }
-      digit = base64.decode(aStr.charAt(i++));
+      digit = base64.decode(aStr.charAt(aIndex++));
       continuation = !!(digit & VLQ_CONTINUATION_BIT);
       digit &= VLQ_BASE_MASK;
       result = result + (digit << shift);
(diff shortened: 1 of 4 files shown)


fitzgen:
(In the future, could you do two PRs, so that we can discuss the diffs separately. Thanks!)

Regarding 1: Interesting, I had assumed that because we weren't doing any operations that would flatten a rope string or cause a copy of a dependent string, the slicing shouldn't have any overhead. Do we actually unknowingly trigger a flattening? Or is it the bookkeeping overhead associated with creating new ropes/dependent strings that point into the shared structure?

fitzgen:
In any case, r+ for part 1.

fitzgen:
Tentative r+ for part 2, meaning we can land this now, but I'd like to explore some follow ups.

I'd like to experiment with moving the cache layer down into parsing individual base 64 VLQs instead of mappings. This will let us avoid allocating a bunch of arrays inside the (very hot) parseMappings loop. Furthermore, we could even generate the cache for base 64 VLQs at "compile time" aka just hard code it in so that we don't have to fill the cache at runtime. Most VLQs are small numbers because each mapping segment is relative, so we should be able to get a huge amount of bang for our buck with just caching the first ~10-20 natural numbers.

ejpbruel:
In response to your first comment, the slicing itself doesn't have any overhead, because we create a JSDependentString internally that does not copy string contents. But there is still the overhead involved in creating a new JSString object, which is small, but apparently negligible.

With regard to your second comment, my original attempt put the cache layer on the level of individual base 64 VLQs, but this actually caused performance to *decrease* in my measurements. My theory is that there is some overhead involved in looking up a VLQ in the cache, so this is only worthwhile if we have a cache hit most of the time *and* the original operation was more expensive. However, the vast majority of VLQs seem to be only a single letter long. Apparently, parsing a single letter VLQ is not expensive enough compared to the overhead of adding a cache lookup to make the latter worthwhile.

The idea behind moving the cache layer to the segment levels is that segments are at least 4 letters long. Parsing a 4 letter segment should be expensive enough to make it worthwhile to cache, provided the level of homogeneity between segments is high enough (which seems to be the case, since I measured a 5% performance increase on my machine).

PR #160, opened 2015-02-20
Rewrite the mappings parser in asm.js

fitzgen:
Still a WIP, but this was pretty fun. Now I need sleep.

With this patch:

Samples 	| Total (ms) |	Mean (ms) |	Standard Deviation (%)
------|---------------------|------------------|---------------------------------
42 |	20023.873 |	476.759 |	7.60

Without this patch:

Samples 	|Total (ms) |	Mean (ms) |	Standard Deviation (%)
-------|-----------|-------------|--------------
1 |	59212.792 |	59212.792 |	NaN

Yes you are reading those numbers correctly. To be fair, the asm.js version doesn't correctly handle names and sources offsets yet, some latent bug that I'm too sleepy to figure out right now. Still.

(NaN is because samples = 1)

### DIFF
--- a/lib/source-map/asm-parser.js
+++ b/lib/source-map/asm-parser.js
@@ -1,0 +1,402 @@
+/* -*- Mode: js; js-indent-level: 2; -*- */
+/*
+ * Copyright 2011 Mozilla Foundation and contributors
+ * Licensed under the New BSD license. See LICENSE or:
+ * http://opensource.org/licenses/BSD-3-Clause
+ */
+if (typeof define !== 'function') {
+    var define = require('amdefine')(module, require);
+}
+define(function (require, exports, module) {
+
+  var util = require('./util');
+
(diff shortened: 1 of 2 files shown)


past:
Whoa!

ai:
Cool!

ejpbruel:
BOOM!

MikeRatcliffe:
Wow, maybe we should write all our tools using asm.js... well, probably just the prettification code ;)

Seriously though:
59212.792 / 476.759 = 124.2
124.2 * 42 = 5216x speed increase... can it really be that much faster?

fitzgen:
Aaaaaand this patch is broken so that the mappings weren't actually parsed and the benchmark doesn't test whether or not the mappings were really parsed.

Sometimes when it looks to good to be true, it actually is too good to be true.

PR #162, opened 2015-02-24
originalPositionFor for should have a GREATEST_LOWER_BOUND search bias.

ejpbruel:
As we discussed per e-mail, we actually only want a LEAST_UPPER_BOUND search bias for generatedPositionOf, not originalPositionOf, so this patch reverts the change in behavior introduced by the earlier patch.


ejpbruel:
I'm closing this pull request until we've sorted out the correct approach for this in https://github.com/mozilla/source-map/pull/154

PR #163, opened 2015-02-24
Always return the smallest element when there is more than one match.

ejpbruel:
The title pretty much sums it up. As discussed per e-mail.


fitzgen:
I want to hold off on further changes here till we fix this mess: https://github.com/mozilla/source-map/pull/154#issuecomment-75791087

PR #165, opened 2015-02-25
Add an optional bias to binary search.

ejpbruel:
This reintroduces the optional bias to the binary search algorithm, but -unlike the previous time- without changing the default bias.

Note that this doesn't add the bias parameter to original/generatedPositionFor yet either (the next pull request will do that), so the observable behavior of the source-map library is still unchanged by this.


sokra:
hmm... you better rebase this, to make the revert commit obsolte...

PR #168, opened 2015-02-26
Always return the smallest element when there is more than one match.

ejpbruel:
I forgot to put this pull request up again yesterday. This changes the search algorithm so that if there is more than one (inexact) match, it always returns the smallest one.

I'm going to merge the current version of the source-map library, with this included, to fx-team.

### DIFF
--- a/lib/source-map/binary-search.js
+++ b/lib/source-map/binary-search.js
@@ -95,8 +95,23 @@ define(function (require, exports, module) {
       return -1;
     }
 
-    return recursiveSearch(-1, aHaystack.length, aNeedle, aHaystack, aCompare,
-                           aBias || exports.GREATEST_LOWER_BOUND);
+    var index = recursiveSearch(-1, aHaystack.length, aNeedle, aHaystack,
+                                aCompare, aBias || exports.GREATEST_LOWER_BOUND);
+    if (index < 0) {
+      return -1;
+    }
+
+    // We have found either the exact element, or the next-closest element than
+    // the one we are searching for. However, there may be more than one such
+    // element. Make sure we always return the smallest of these.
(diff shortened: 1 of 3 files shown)


lib/source-map/binary-search.js
```diff
@@ -95,8 +95,23 @@ define(function (require, exports, module) {
...
+                                aCompare, aBias || exports.GREATEST_LOWER_BOUND);
+    if (index < 0) {
+      return -1;
+    }
+
+    // We have found either the exact element, or the next-closest element than
+    // the one we are searching for. However, there may be more than one such
+    // element. Make sure we always return the smallest of these.
+    while (index - 1 > 0) {
+      if (aCompare(aHaystack[index], aHaystack[index - 1], true) !== 0) {
```
fitzgen: Why `index - 1 > 0`? Shouldn't it be `index - 1 >= 0`? Otherwise, won't you not find the first match if the very first item in the array is the first one? You'll get the second one in that case, right? If I'm wrong, then we need a comment here explaining this, because it isn't clear to me (and therefore likely not clear to others, as well) from just reading the code.

Also, can we expand the test to include the first-match-is-first-in-array case? Something like this:

```js
var needle = 1;
var haystack = [1, 1, 1, 2, 5, 5, 5, 13, 21];
assert.equal(binarySearch.search(needle, haystack, numberCompare,
                                 binarySearch.LEAST_UPPER_BOUND), 
             0);
```

fitzgen: Or alternatively `index > 0`.

ejpbruel:
Updated the PR to address your comments.

fitzgen:
Thanks!

PR #169, opened 2015-02-27
Add "files" array to package.json

hawkrives:
Currently, this module is 15MB:

```
 15M	./bench
 40K	./build
112K	./lib
132K	./test
 16M	total
```

npm has a [`files` whitelist](https://docs.npmjs.com/files/package.json#files), which activates when publishing to npm. By excluding some files that aren't used (or cannot be used) by packages that depend on this module, we can shrink the required size drastically:

```
112K	./lib
156K	total
```

I have set `files` to include `lib/`, `LICENSE`, and `README`.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+bench/
+test/


fitzgen:
It seems like the only problem is the `bench` subdirectory. Would prefer to maintain a blacklist rather than a whitelist so that we don't have to remember to add new files to the whitelist if they are outside of `lib`.

Can you just do a `.npmignore` file that lists the `bench` directory instead? Thanks.

hawkrives:
If you opt for that, I'd request that the `test` directory is also ignored.

hawkrives:
Although I'd also be quite happy with just ignoring the test dir.

fitzgen:
> If you opt for that, I'd request that the `test` directory is also ignored.

Sure.

hawkrives:
I assume you would prefer to keep the build/ directory in the npm package?

rubennorte:
Please, merge this PR so we can save a lot of time and space installing this package ;)

fitzgen:
Thanks!

zertosh:
@fitzgen Please publish this to npm :smile:

PR #170, opened 2015-02-27
fixed global leakage of variable `value`

sokra:


sokra:
Any reason why not using a test runner like mocha?

fitzgen:
> Any reason why not using a test runner like mocha?

The tests need to run inside Firefox's internal test suite as well as on node, and at the time writing our own test runner was the easiest option.

fitzgen:
Thanks

sokra:
Could you please publish this?

fitzgen:
Published.

sokra:
Thanks

PR #171, opened 2015-03-05
When finding generated positions, mappings should end at the end of the source

ejpbruel:
I ran into this when implementing the new sliding breakpoint algorithm for source mapped sources. I assumed that generatedPositionOf returns a null mapping if there is no closest mapping that is greater than the one we are searching for *for the same original source*. If we use generatedPositionOf with a least upper bound bias, this means we get a null mapping if and only if we went past the last line in the original source, in which case the breakpoint sliding algorithm failed.

The problem is that when we have two original sources A and B, such that A comes lexicographically before B, all mappings for B are considered to be greater than A, so that when we go past the last line in A, generatedPositionOf returns the first mapping for B instead of a null mapping.

### DIFF
--- a/lib/source-map/source-map-consumer.js
+++ b/lib/source-map/source-map-consumer.js
@@ -641,11 +641,13 @@ define(function (require, exports, module) {
       if (index >= 0) {
         var mapping = this._originalMappings[index];
 
-        return {
-          line: util.getArg(mapping, 'generatedLine', null),
-          column: util.getArg(mapping, 'generatedColumn', null),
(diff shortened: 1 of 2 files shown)


ejpbruel:
I just ran into another issue where, when you search for an original location with an absolute URL, this patch doesn't handle source roots correctly. Please wait with reviewing this until I update the PR.

ejpbruel:
Turns out I didn't really understand source roots, and the problem was not where I thought it was. This PR is fine as it is, so feel free to review it. I'll file a separate pull request for the actual issue.

fitzgen:
LGTM!

PR #172, opened 2015-03-06
 allGeneratedPositionsFor for line with no mappings should return the next closest line.

ejpbruel:
We assume that allGeneratedPositionsFor returns an empty list if and only if we've reached the end of the original source: that is, there are no lines greater than or equal than the one we are searching for that have any mappings. We use this in breakpoint sliding for source-mapped sources to determine whether we've reached the end of the original source.

The problem is that allGeneratedPositionsFor also returns an empty lists if there are no mappings for the exact line we are looking for. What it should do is simply return the list of all mappings for the next closest line, if any. This is also consistent with how generatedPositionFor works (when using a LUB bias).

### DIFF
--- a/lib/source-map/source-map-consumer.js
+++ b/lib/source-map/source-map-consumer.js
@@ -201,11 +201,12 @@ define(function (require, exports, module) {
                                     binarySearch.LEAST_UPPER_BOUND);
       if (index >= 0) {
         var mapping = this._originalMappings[index];
+        var originalLine = mapping.originalLine;
 
         // Iterate until either we run out of mappings, or we run into
         // a mapping for a different line. Since mappings are sorted, this is
         // guaranteed to find all mappings for the line we are searching for.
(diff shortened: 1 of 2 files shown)


lib/source-map/source-map-consumer.js
```diff
@@ -201,11 +201,12 @@ define(function (require, exports, module) {
                                     binarySearch.LEAST_UPPER_BOUND);
       if (index >= 0) {
         var mapping = this._originalMappings[index];
+        var originalLine = mapping.originalLine;
 
         // Iterate until either we run out of mappings, or we run into
         // a mapping for a different line. Since mappings are sorted, this is
         // guaranteed to find all mappings for the line we are searching for.
-        while (mapping && mapping.originalLine === needle.originalLine) {
+        while (mapping && mapping.originalLine === originalLine) {
```
fitzgen: Good find!

fitzgen:
Will rebase + merge.

PR #173, opened 2015-03-11
Add a column argument to allGeneratedPositionsFor.

ejpbruel:
Some source maps define multiple mappings for a single original column. We need
a way to obtain all mappings for a given column, instead of just the first one.
This patch refactors allGeneratedPositionsFor so that it returns all mappings
for a single line if no column argument is passed (not necessarily the line that
was requested, due to fuzzy search), and all mappings for a single line and
column otherwise.

### DIFF
--- a/test/source-map/test-source-map-consumer.js
+++ b/test/source-map/test-source-map-consumer.js
@@ -495,7 +495,7 @@ define(function (require, exports, module) {
     assert.equal(pos.column, 2);
   };
 
-  exports['test allGeneratedPositionsFor'] = function (assert, util) {
+  exports['test allGeneratedPositionsFor for line'] = function (assert, util) {
     var map = new SourceMapGenerator({
       file: 'generated.js'
     });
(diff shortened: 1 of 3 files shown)


test/source-map/test-source-map-consumer.js
```diff
@@ -583,6 +583,34 @@ define(function (require, exports, module) {
     assert.equal(mappings.length, 0);
   };
 
+  exports['test allGeneratedPositionsFor for position with multiple mappings'] = function (assert, util) {
```
fitzgen: This adds tests for multiple mappings for a single generated position, however there are no tests for passing a column into `allGeneratedPositionsFor`. Please add tests for that.

Also, how is the column parameter expected to interact with the fuzzy searching? Please add a test for this case as well.

fitzgen:
Looks good overall, but see inline comment above.

ejpbruel:
Oops. That test was supposed to be testing adding a column to allGeneratedPositions for. I've updated the test, and also added one with fuzzy search.

fitzgen:
Thanks!

PR #176, opened 2015-04-02
Skipping mappings that exceed line count in generated code

### LINKED ISSUE #174: SourceNode.fromStringWithSourceMap throws when there are more mapping lines than generated code lines
This was reported in https://github.com/mozilla/source-map/issues/138 but fixed by a bug in another project. I think it is still a valid source-map bug as [stated here](https://github.com/mozilla/source-map/issues/138#issuecomment-62853402).

The scenario I have is:

1. Source map is created from escodegen.
2. Source map and code from 1 is ran through uglify which generates a new source map.
3. Calling SourceNode.fromStringWithSourceMap with the code/map from 2 will throw.

The reason it throws is that uglify calls `addMapping` [here](https://github.com/mishoo/UglifyJS2/blob/master/lib/sourcemap.js#L80) but the existing mappings remain from the original transformation.

fromStringWithSourceMap expects there to be fewer generatedLines than lines in the generated code.

I'd be happy to submit a patch for this but could use some guidance on what needs to be done to fix this.

matthewp:
This is a fix for #174. In `fromStringWithSourceMap` we should not
attempt to create mappings for the SourceNode when there are mappings
for lines that exceed the number of lines in the generated code. This
can happen in scenarios where a sourcemap has been transformed a number
of times and new mappings are added each time. There will be "dead
mappings" that refer to generatedLines that no longer exceed in the
generated code. We can safely ignore these. Fixes #174

### DIFF
--- a/lib/source-map/source-node.js
+++ b/lib/source-map/source-node.js
@@ -66,6 +66,7 @@ define(function (require, exports, module) {
       // (since `REGEX_NEWLINE` captures its match).
       // Processed fragments are removed from this array, by calling `shiftNextLine`.
       var remainingLines = aGeneratedCode.split(REGEX_NEWLINE);
+      var lineCount = remainingLines.length;
       var shiftNextLine = function() {
         var lineContents = remainingLines.shift();
         // The last line of a file might not have a newline.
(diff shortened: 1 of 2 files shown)


fitzgen:
See comment here: https://github.com/bitovi/source-map/commit/377291d28456eb77ed8dc1dbd432330ab7165053#commitcomment-10565532

This is a bug in code using the source-map library, not a bug in the library itself.

PR #177, opened 2015-04-03
allGeneratedPositionsFor should only return mappings for the exact line ...

ejpbruel:
...when a column is specified.

This is a small bug that I uncovered during testing of the breakpoint sliding algorithm.


fitzgen:
Great bug find. Thanks!

PR #180, opened 2015-05-03
Use files in package.json

SimenB:
http://hastebin.com/ibumineyag.sh vs. http://hastebin.com/xutifitico.sh

Difference is dotfiles and `Makefile.dryice.js`.

I'm unsure if `build/` is actually needed for npm, but as it's included now, I left it in :smile:

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,0 @@
-bench/
-test/
--- a/package.json
+++ b/package.json
@@ -46,6 +46,10 @@
     "lib": "./lib"
   },
   "main": "./lib/source-map.js",
+  "files": [
+    "lib/",
+    "build/"
+  ],
   "engines": {
     "node": ">=0.8.0"
   },


fitzgen:
Somehow this PR got lost in my inbox, sorry for the delayed response.

I don't think this is necessary anymore because we use .npmignore to ignore the couple large files that are not needed by projects depending on this one.

If I'm mistaken and this would still provide further value, please explain and reopen.

SimenB:
It's mostly a blacklist vs whitelist thing.

If you add a new file for development purposes, maybe a gulpfile or another dotfile, you would now have to add it to `npmignore` to avoid including it in the package. Using `files` instead it would stay out. And if you add another directory, you'd have to explicitly include it (unless it's in `build` or `lib`).

Some files are included even though they are not in `files` such as `readme`, `changelog` and `license`. And of course `package.json`.

SimenB:
I'll do screenshots of the diff, seeing as hastebin deletes old pastes.

Current:
image

If you merge this:
image

fitzgen:
Ok, that sounds fine to me. Can you rm .npmignore as well? Will merge if you do that.

SimenB:
It's been removed already :smile:

fitzgen:
Ah, thanks! Waiting for the new travis-ci run.

PR #182, opened 2015-05-05
Fix util.relative for the case where the path is above the root.

ejpbruel:
Currently, util.relative simply checks if the root is a prefix of the path. This
works fine when the path is below the root, but won't work if the path is above
the root.

To fix this, we should remove components from the root one by one, until either
we find a prefix that fits, or we run out of components to match.

### DIFF
--- a/lib/source-map/util.js
+++ b/lib/source-map/util.js
@@ -198,15 +198,30 @@ define(function (require, exports, module) {
 
     aRoot = aRoot.replace(/\/$/, '');
 
-    // XXX: It is possible to remove this block, and the tests still pass!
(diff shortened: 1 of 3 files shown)


ejpbruel:
I understand that fitzgen is currently on vacation. Is there anyone who can merge this pull request in his absence?

ejpbruel:
@lydell could you review this pull request? I would love to fix this bug this week, but am reluctant to just merge it without review.

lydell:
A few more tests of `util.relative` as well as a test using the public API would be great.

ejpbruel:
Added the tests you requested.

PR #185, opened 2015-06-17
Optimize the base64.decode function.

fitzgen:
This function is incredibly hot while parsing mappings. By avoiding property
gets and throwing errors in this function, we get about 2000 millisecond
improvement on the mean time to parse our source map in `bench.html`.

r? @ejpbruel

### DIFF
--- a/lib/source-map/base64.js
+++ b/lib/source-map/base64.js
@@ -9,34 +9,65 @@ if (typeof define !== 'function') {
 }
 define(function (require, exports, module) {
 
-  var charToIntMap = {};
-  var intToCharMap = {};
-
-  'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
-    .split('')
-    .forEach(function (ch, index) {
-      charToIntMap[ch] = index;
-      intToCharMap[index] = ch;
-    });
+  var intToCharMap = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'.split('');
 
   /**
    * Encode an integer in the range of 0 to 63 to a single base 64 digit.
    */
-  exports.encode = function base64_encode(aNumber) {
-    if (aNumber in intToCharMap) {
-      return intToCharMap[aNumber];
+  exports.encode = function (number) {
+    if (0 <= number && number < intToCharMap.length) {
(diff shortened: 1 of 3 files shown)


ai:
BTW, why you didn’t use base64 function from some npm library?

fitzgen:
At the time this lib was originally written, I'm not sure there was a good solution on npm. It was easy to implement it myself.

The requirement of running on node, in browsers, and within Firefox is another consideration, but not a show stopper.

lib/source-map/base64.js
```diff
@@ -9,34 +9,62 @@ if (typeof define !== 'function') {
...
   /**
-   * Decode a single base 64 digit to an integer.
+   * Decode a single base 64 character code digit to an integer. Returns -1 on
+   * failure.
    */
-  exports.decode = function base64_decode(aChar) {
-    if (aChar in charToIntMap) {
-      return charToIntMap[aChar];
+  exports.decode = function (charCode) {
+    var bigA = 65;     // 'A'
```
ejpbruel: Is the JIT clever enough not to reinitialize these variables on each call to decode? If not, these constants should be moved outside the function

lib/source-map/base64.js
```diff
@@ -9,34 +9,62 @@ if (typeof define !== 'function') {
...
+    var slash = 47;    // '/'
+
+    // 0 - 25: ABCDEFGHIJKLMNOPQRSTUVWXYZ
+    if (bigA <= charCode && charCode <= bigZ) {
+      return (charCode - bigA);
+    }
+
+    // 26 - 51: abcdefghijklmnopqrstuvwxyz
+    if (littleA <= charCode && charCode <= littleZ) {
+      return (charCode - littleA + 26);
```
ejpbruel: You introduced several named variables for magic constants above, only to use an unnamed magic number here. Would be nice if we could come up with a name for these as well.

ejpbruel:
Patch looks good to me. r+

It's a bit surprising that a property lookup is so much slower than a bunch of hardcoded if blocks. Intuitively, I'd expect table lookup to always win out over explicit branching (when not taking cache effects into account).

In any case, I wonder if doing a lookup based on the character's charCode, rather than it's string representation, would be any faster. Array element lookup is supposed to be faster than object property lookup, so I guess it depends on what the JIT can optimize. I expect it won't be faster than the inline code you just wrote, but it might be worth looking into at some point.

PR #186, opened 2015-06-18
Variety of speed ups for parsing mappings

fitzgen:
This builds on top of https://github.com/mozilla/source-map/pull/185

This is a variety of speed ups to parsing mappings. Most notable is rolling our own sorting in JS. See https://github.com/mozilla/source-map/blob/a0f0ff88a4c3f29023acc24118493703684cae84/lib/source-map/quick-sort.js#L12-L20 for a description of why that is.

r? @ejpbruel

### DIFF
--- a/bench/bench.js
+++ b/bench/bench.js
@@ -10,7 +10,7 @@ function benchmark(name, setup, action) {
 
   // Warm up the JIT.
   var start = Date.now();
-  while ((Date.now() - start) < 10000 /* 10 seconds */) {
+  while ((Date.now() - start) < 5000 /* 5 seconds */) {
     action();
   }
 
@@ -18,7 +18,7 @@ function benchmark(name, setup, action) {
 
   console.profile(name);
   var start = Date.now();
-  while ((Date.now() - start) < 20000 /* 20 seconds */) {
+  while ((Date.now() - start) < 30000 /* 30 seconds */) {
     var thisIterationStart = window.performance.now();
     action();
     stats.take(window.performance.now() - thisIterationStart);
--- a/lib/source-map/quick-sort.js
+++ b/lib/source-map/quick-sort.js
@@ -1,0 +1,128 @@
+/* -*- Mode: js; js-indent-level: 2; -*- */
+/*
+ * Copyright 2011 Mozilla Foundation and contributors
+ * Licensed under the New BSD license. See LICENSE or:
+ * http://opensource.org/licenses/BSD-3-Clause
+ */
+if (typeof define !== 'function') {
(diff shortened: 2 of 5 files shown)


bench/bench.js
```diff
@@ -10,7 +10,7 @@ function benchmark(name, setup, action) {
 
   // Warm up the JIT.
   var start = Date.now();
-  while ((Date.now() - start) < 10000 /* 10 seconds */) {
+  while ((Date.now() - start) < 5000 /* 5 seconds */) {
```
ejpbruel: This seems like it would be race condition prone, since the results are timing dependent. Is there a better way to warm up the JIT, that does not suffer from timing effects?

lib/source-map/quick-sort.js
```diff
@@ -0,0 +1,128 @@
...
+   * @param {Array} ary
+   *        An array to sort.
+   * @param {function} comparator
+   *        Function to use to compare two items.
+   * @param {Number} p
+   *        Start index of the array
+   * @param {Number} r
+   *        End index of the array
+   */
+  function doQuickSort(ary, comparator, p, r) {
```
ejpbruel: Function calls are supposed to be quite expensive in JavaScript. I wonder if we could do even better here by using a non-recursive merge sort, or a similar non-recursive sorting algorithm.

lib/source-map/quick-sort.js
```diff
@@ -0,0 +1,128 @@
...
+   * @param {Array} ary
+   *        The array to partition.
+   * @param {function} comparator
+   *        The function with which to compare items.
+   * @param {Number} p
+   *        Start index of the array.
+   * @param {Number} r
+   *        End index of the array.
+   */
+  function partition(ary, comparator, p, r) {
```
ejpbruel: Why don't you inline this function? It seems to be only used from one place. We could save the overhead of a single function call for each recursive step.

ejpbruel:
Patch looks good to me. r+

These are great improvements. It's quite fun how we're scraping the barrel of JS performance with this library :-)

bench/bench.js
fitzgen: Unfortunately, because each parse takes so long, we can't just run it n times before measuring, and I don't think there is a better way.

lib/source-map/quick-sort.js
fitzgen: Yes, and the jit coach is pointing out that these calls aren't getting inlined, and for some reason doQuickSort often stays in baseline. Very surprising given the performance gains it still manages to provide! Definitely a place to investigate more.

PR #188, opened 2015-06-26
More jit profiling

fitzgen:
r? @ejpbruel 

A couple code clean ups and also some more big perf gains parsing source maps.

Notable gains are

* Helping TI out with a constructor for mappings

* Not eagerly inflating source names, which lets us compare source indices rather than strings when comparing mappings.


ejpbruel:
This patch looks good to me. Good catch Nick!

PR #190, opened 2015-07-07
Add SourceMapConsumer.prototype.hasContentsOfAllSources

fitzgen:
r? @jlongster 

Part 1 of fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1166048

### DIFF
--- a/build/prefix-source-map.jsm
+++ b/build/prefix-source-map.jsm
@@ -17,4 +17,5 @@
 
 this.EXPORTED_SYMBOLS = [ "SourceMapConsumer", "SourceMapGenerator", "SourceNode" ];
 
+Components.utils.import("resource://gre/modules/devtools/Console.jsm");
 Components.utils.import('resource://gre/modules/devtools/Require.jsm');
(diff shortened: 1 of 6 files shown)


build/prefix-source-map.jsm
```diff
@@ -17,4 +17,5 @@
 
 this.EXPORTED_SYMBOLS = [ "SourceMapConsumer", "SourceMapGenerator", "SourceNode" ];
 
+Components.utils.import("resource://gre/modules/devtools/Console.jsm");
```
jsantell: extra?

jlongster:
r+ looks good

build/prefix-source-map.jsm
jlongster: noticed that, but I thought we were ok leaving those in, it will only ever be imported once and it's useful for future debugging?

fitzgen: Yeah left it in on purpose so we don't have to add it again when debugging which is annoying.

PR #193, opened 2015-07-09
Do mutate shared state with the old SourceMapGenerator after calling …

### LINKED ISSUE #191: Error in 0.4.3: "0" is not in the set.
I update source-map in PostCSS to 0.4.3 and got a errors in tests on `map.toString()`

```
Error: "0" is not in the set.
```

/cc @fitzgen

fitzgen:
…SourceMapConsumer.fromSourceMap.

Fixes #191.

r? @jlongster

### DIFF
--- a/lib/source-map/source-map-consumer.js
+++ b/lib/source-map/source-map-consumer.js
@@ -337,25 +337,45 @@ define(function (require, exports, module) {
     function SourceMapConsumer_fromSourceMap(aSourceMap) {
       var smc = Object.create(BasicSourceMapConsumer.prototype);
 
-      smc._names = ArraySet.fromArray(aSourceMap._names.toArray(), true);
-      smc._sources = ArraySet.fromArray(aSourceMap._sources.toArray(), true);
+      var names = smc._names = ArraySet.fromArray(aSourceMap._names.toArray(), true);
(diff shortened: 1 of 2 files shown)


fitzgen:
Woops, left a require('...') call in there that shouldn't have been in there. Force pushed a fix.

lib/source-map/source-map-consumer.js
```diff
@@ -337,25 +337,45 @@ define(function (require, exports, module) {
...
 
-          if (m.name !== null) {
-            m.name = smc._names.indexOf(m.name);
+      for (var i = 0, length = generatedMappings.length; i < length; i++) {
+        var srcMapping = generatedMappings[i];
+        var destMapping = new Mapping;
+        destMapping.generatedLine = srcMapping.generatedLine;
+        destMapping.generatedColumn = srcMapping.generatedColumn;
+
+        if (srcMapping.source) {
```
jlongster: Any reason you dropped the explicit `!== null` checks? I feel like I remember you adding them somewhere else for JIT reasons, I could be completely wrong though.

jlongster:
I wonder if this will make pretty printing slower, but it seems like the right thing to do so r+

fitzgen:
Just fixed up the commit message.

Issue #192, opened 2015-07-08
Wrong column number in 0.4.3


iamvdo:
As #191, I also get errors with 0.4.3 from PostCSS sourcemap.

It returns wrong column from `generatedPositionFor`.

I have this file `module.css`:
```css
.module {
  color: #fff
}
```
And the processed file:
```css
.import{color:#fff}.module{color:#fff}.main{color:#fff}
```

When I use :
```javascript
var smc = new SourceMap.SourceMapConsumer(PostCSS_map);
var positions = smc.generatedPositionFor({source: 'module.css', line: 1, column: 0});
```
`positions.column == 19` with @0.4.2 and `positions.column == 27` with @0.4.3

Is it intended? Am I missing something?

fitzgen:
Can you package up the test case into a gist and give me steps to reproduce with the packaged test case? That will help me debug the issue. Thanks!

iamvdo:
Well, here you go: https://gist.github.com/iamvdo/7051563e4383dee93da8

Install modules
```
npm i postcss csswring source-map
```

Then switch between `source-map@0.4.2` and `source-map@0.4.3` from `node_modules/postcss` folder.

With `source-map@0.4.2`, you get mappings for `module.css`:
```
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 19,
  originalLine: 1,
  originalColumn: 0,
  name: null }
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 27,
  originalLine: 2,
  originalColumn: 2,
  name: null }
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 37,
  originalLine: 2,
  originalColumn: 14,
  name: null }
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 38,
  originalLine: 3,
  originalColumn: 1,
  name: null }
```
While with `source-map@0.4.3`, you get:
```
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 27,
  originalLine: 2,
  originalColumn: 2,
  name: null }
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 37,
  originalLine: 2,
  originalColumn: 14,
  name: null }
{ source: 'module.css',
  generatedLine: 1,
  generatedColumn: 38,
  originalLine: 3,
  originalColumn: 1,
  name: null }
```

[...]

PR #194, opened 2015-07-09
Fix for #192

fitzgen:
r? @jlongser 

Ok so in the process of moving to deflated mappings for source map parsing perf gains, the source map generating was now using a compare function that expected `.name` and `.source` to be indices rather than strings.

/me mumbles something about types and having a compiler find bugs for you

This PR splits `compareByGeneratedPosition` into two versions: one for inflated mappings and one for deflated mappings, and adds a regression test.

As I mentioned in #191, we should probably move `SourceMapGenerator` to using deflated mappings as well, since that will likely give a nice perf boost and also clean up these edge cases.

### DIFF
--- a/lib/source-map/mapping-list.js
+++ b/lib/source-map/mapping-list.js
@@ -22,7 +22,7 @@ define(function (require, exports, module) {
     var columnA = mappingA.generatedColumn;
     var columnB = mappingB.generatedColumn;
     return lineB > lineA || lineB == lineA && columnB >= columnA ||
-           util.compareByGeneratedPositions(mappingA, mappingB) <= 0;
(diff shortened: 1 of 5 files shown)


fitzgen:
Going to rebase on master so the diff isn't so cluttered.

fitzgen:
> As I mentioned in #191, we should probably move SourceMapGenerator to using deflated mappings as well, since that will likely give a nice perf boost and also clean up these edge cases.

Filed #195 for this.

lydell:
If you meant to ping @jlongster you missed a t in the OP :)

fitzgen:
Thanks. Yes, meant to r? @jlongster

jlongster:
r+ looks good, had to figure out what deflated/inflated was but I see now.

PR #198, opened 2015-07-29
Bug 1182194 - Tweak in order to use source-map.js instead of SourceMap.jsm. r=fitzgen

ochameau:
Two small changes.
I'm getting rid of Require.jsm usage on mozilla-central. Instead I'm using builtin module loading shipped within source-map.js and its module loader doesn't take care of `define` second argument `['exports`]`, it just always pass `require` and `exports`.

More context in the bug https://bugzilla.mozilla.org/show_bug.cgi?id=935366

### DIFF
--- a/build/suffix-browser.js
+++ b/build/suffix-browser.js
@@ -6,3 +6,6 @@ this.sourceMap = {
   SourceMapGenerator: require('source-map/source-map-generator').SourceMapGenerator,
   SourceNode: require('source-map/source-node').SourceNode
 };
+if (typeof module === "object" && module && module.exports) {
+  module.exports = this.sourceMap;
+}
--- a/test/source-map/test-api.js
+++ b/test/source-map/test-api.js
@@ -13,8 +13,8 @@ define(function (require, exports, module) {
   try {
     sourceMap = require('../../lib/source-map');
   } catch (e) {
-    sourceMap = {};
-    Components.utils.import('resource:///modules/devtools/SourceMap.jsm', sourceMap);
+    var loader = Components.utils.import("resource://gre/modules/devtools/Loader.jsm", {});
+    sourceMap = loader.devtools.require("devtools/toolkit/sourcemap/source-map");
   }
 
   exports['test that the api is properly exposed in the top level'] = function (assert, util) {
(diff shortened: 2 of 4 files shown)


build/suffix-browser.js
```diff
@@ -6,3 +6,6 @@ this.sourceMap = {
   SourceMapGenerator: require('source-map/source-map-generator').SourceMapGenerator,
   SourceNode: require('source-map/source-node').SourceNode
 };
+if (module && module.exports) {
```
fitzgen: This will break on the web, need to guard against the `module` binding not existing at all:

    if (typeof module === "object" && module && module.exports)

fitzgen:
LGTM, thanks!

Can you verify that the browser build still works with your changes? `node Makefile.dryice.js` and then run the benchmark. https://github.com/mozilla/source-map/blob/master/bench/README.md

Feel free to merge it in after fixing up that one nit and verifying the browser build still works.

ochameau:
Benchmark with this patch:
```
Parsing source map
[Stats samples: 8, total: 24379 ms, mean: 3047.375 ms, stddev: 162.02790135136146 ms]

Serializing source map
[Stats samples: 2, total: 26248 ms, mean: 13124 ms, stddev: 86 ms]
```

Without this patch:
```
Parsing source map
[Stats samples: 8, total: 24527 ms, mean: 3065.875 ms, stddev: 139.97835109583983 ms]

Serializing source map
[Stats samples: 2, total: 27254 ms, mean: 13627 ms, stddev: 416 ms]
```

Try run:
  https://treeherder.mozilla.org/#/jobs?repo=try&revision=56770ce87784

test/source-map/test-api.js
```diff
@@ -13,8 +13,8 @@ define(function (require, exports, module) {
   try {
     sourceMap = require('../../lib/source-map');
   } catch (e) {
-    sourceMap = {};
-    Components.utils.import('resource:///modules/devtools/SourceMap.jsm', sourceMap);
+    var {devtools} = Components.utils.import("resource://gre/modules/devtools/Loader.jsm", {});
+    sourceMap = devtools.require("devtools/toolkit/sourcemap/source-map");
```
fitzgen: These tests have to be able to run in node, so no ES6 :(

https://travis-ci.org/mozilla/source-map/jobs/73406398#L107

fitzgen:
Thanks. Note that Github doesn't send emails unless you comment, so I had no idea you had fixed this PR up.

PR #199, opened 2015-07-29
Bug in getting embedded source contents

fitzgen:
This is just a failing test case right now. Haven't dug into it yet.

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1188697

### DIFF
--- a/lib/source-map/source-map-consumer.js
+++ b/lib/source-map/source-map-consumer.js
@@ -305,10 +305,20 @@ define(function (require, exports, module) {
       throw new Error('Unsupported version: ' + version);
     }
 
-    // Some source maps produce relative source paths like "./foo.js" instead of
-    // "foo.js".  Normalize these first so that future comparisons will succeed.
-    // See bugzil.la/1090768.
-    sources = sources.map(util.normalize);
+    sources = sources
+      // Some source maps produce relative source paths like "./foo.js" instead of
+      // "foo.js".  Normalize these first so that future comparisons will succeed.
+      // See bugzil.la/1090768.
+      .map(util.normalize)
+      // Always ensure that absolute sources are internally stored relative to
+      // the source root, if the source root is absolute. Not doing this would
+      // be particularly problematic when the source root is a prefix of the
+      // source (valid, but why??). See github issue #199 and bugzil.la/1188982.
+      .map(function (source) {
+        return sourceRoot && util.isAbsolute(sourceRoot) && util.isAbsolute(source)
+          ? util.relative(sourceRoot, source)
+          : source;
+      });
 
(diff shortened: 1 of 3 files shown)


fitzgen:
Ok, figured this out. We rely on maintaining the invariant that sources are internally stored relative to the source root (except when the source is absolute and the source root is not) and there is a case where we weren't maintaining this: when the source is absolute and the source root is a prefix of the source. Why would a tool do that? No idea. Probably not on purpose.

r? @jryans

fitzgen:
Oops jryans is on pto. I just chose him because he had dealt with source root issues before but it doesn't need his review in particular.

r? @ejpbruel

fitzgen:
Everyone is on PTO...

r? @tromey

lib/source-map/source-map-consumer.js
```diff
@@ -305,10 +305,20 @@ define(function (require, exports, module) {
...
+      // Some source maps produce relative source paths like "./foo.js" instead of
+      // "foo.js".  Normalize these first so that future comparisons will succeed.
+      // See bugzil.la/1090768.
+      .map(util.normalize)
+      // Always ensure that absolute sources are internally stored relative to
+      // the source root, if the source root is absolute. Not doing this would
+      // be particularly problematic when the source root is a prefix of the
+      // source (valid, but why??). See github issue #199 and bugzil.la/1188982.
+      .map(function (source) {
+        return sourceRoot && sourceRoot.charAt(0) === '/' && source.charAt(0) === '/'
```
jryans: I thought most `sourceRoot`s were URLs, not simple local paths? Do we need a better check for "is absolute" to cover those too?

fitzgen: I got halfway through writing a comment about why I didn't think so, but then unconvinced myself and finally realized that we should just add test coverage for more edge cases here...

flying-sheep: you surely meant `charAt(1)`, the second time? those are two times the same check.

but even better would be: `sourceRoot && sourceRoot.startsWith('//')`

fitzgen:
Ok added the test for URLs too.

r? @jryans

lib/source-map/util.js
```diff
@@ -185,6 +185,10 @@ define(function (require, exports, module) {
   }
   exports.join = join;
 
+  exports.isAbsolute = function (aPath) {
```
jryans: Should [`normalize`][1] use this new utility instead of having it own similar check?

[1]: https://github.com/fitzgen/source-map/blob/source-content-bug/lib/source-map/util.js#L93

jryans:
r+, looks good overall!

Issue #179, opened 2015-05-01
Include dist builds


niemyjski:
Could you please include dist builds. This would make it easier to consume this library. Also the building of dist files fails under windows:

```
Creating dist/source-map.js
Failed to find module: amdefine from source-map/source-map-generator
Failed to find module: amdefine from source-map/base64-vlq
Failed to find module: amdefine from source-map/base64
Failed to find module: amdefine from source-map/util
Failed to find module: amdefine from source-map/array-set
Failed to find module: amdefine from source-map/mapping-list
Failed to find module: amdefine from source-map/source-map-consumer
Failed to find module: amdefine from source-map/binary-search
Failed to find module: amdefine from source-map/source-node

Creating dist/source-map.min.js
```

mkleehammer:
Similarly fails on OS X

mkleehammer:
According to #140 it is actually succeeding.  However, minified dist builds would be great.

jkrems:
This is likely no longer actionable given that the way the library is built and distributed has changed.

Issue #140, opened 2014-11-18
Remove "Failed to find module: amdefine" spew when successfully running node Makefile.dryice.js


ejpbruel:
I've tried to build the source-map library from source, by following the README.md. When I run the following commands:

npm link
node Makefile.dryice.js

I get the following output:

Creating dist/SourceMap.jsm
Failed to find module: amdefine from source-map/source-map-consumer
Failed to find module: amdefine from source-map/util
Failed to find module: amdefine from source-map/binary-search
Failed to find module: amdefine from source-map/array-set
Failed to find module: amdefine from source-map/base64-vlq
Failed to find module: amdefine from source-map/base64
Failed to find module: amdefine from source-map/source-map-generator
Failed to find module: amdefine from source-map/source-node

Creating dist/test/Utils.jsm
Failed to find module: amdefine from test/source-map/util
Failed to find module: amdefine from lib/source-map/util

Creating dist/test/test_api.js

Creating dist/test/test_array_set.js

Creating dist/test/test_base64_vlq.js

Creating dist/test/test_base64.js

Creating dist/test/test_binary_search.js

Creating dist/test/test_dog_fooding.js

Creating dist/test/test_source_map_consumer.js

Creating dist/test/test_source_map_generator.js

Creating dist/test/test_source_node.js

Creating dist/test/test_util.js

Creating dist/source-map.js
Failed to find module: amdefine from source-map/source-map-generator
Failed to find module: amdefine from source-map/base64-vlq
Failed to find module: amdefine from source-map/base64
Failed to find module: amdefine from source-map/util
Failed to find module: amdefine from source-map/array-set
Failed to find module: amdefine from source-map/source-map-consumer
Failed to find module: amdefine from source-map/binary-search
Failed to find module: amdefine from source-map/source-node

Creating dist/source-map.min.js

I've made sure the amdefine module is installed by running npm -g install amdefine, and that NODE_PATH is set appropriately.

When I create an minimal index.js that simply does require('amdefine'), the module is found without a hitch. Either I'm doing something wrong, or there is something wrong with the build script.

niemyjski:
+1

jkrems:
Closing because the package no longer heavily depends on AMD.

Issue #67, opened 2013-06-28
Stop using AMD modules, use browserify/webpack/commonjs-everywhere instead


duncanbeevers:
See https://github.com/mozilla/source-map/issues/67#issuecomment-20220148

PR #202, opened 2015-08-23
Remove "amdefine" and add `define` during build

### LINKED ISSUE #158: Please add CommonJS-compatible browser build to package.json
Using browserify ends up requiring the `amddefine` module and its dependencies into the build. It would be great if you'd publish the browser build onto npm and add it to the `browser` field in the `package.json` file (see https://github.com/substack/browserify-handbook#browser-field). Thanks!

zertosh:
Fixes #158, #179, #140 and #67.

This PR removes "amdefine" in favor of wrapping the modules in AMD wrappers during build time. This makes "source-map" friendlier to node and browserify/webpack environments – while still providing the AMDness needed for the FF devtools (I think that's why it's there [1](https://bugzilla.mozilla.org/show_bug.cgi?id=935366) [2](https://bugzilla.mozilla.org/show_bug.cgi?id=1134628)). This approach carries no maintenance overhead, and should be 100% compatible with existing uses. 

The problem is that then I own the entire git blame. Any solutions? Maybe leave the indentation alone?

cc: @fitzgen

### DIFF
--- a/Makefile.dryice.js
+++ b/Makefile.dryice.js
@@ -8,23 +8,26 @@ var path = require('path');
 var fs = require('fs');
 var copy = require('dryice').copy;
 
-function removeAmdefine(src) {
-  src = String(src).replace(
-    /if\s*\(typeof\s*define\s*!==\s*'function'\)\s*{\s*var\s*define\s*=\s*require\('amdefine'\)\(module,\s*require\);\s*}\s*/g,
-    '');
-  src = src.replace(
-    /\b(define\(.*)('amdefine',?)/gm,
-    '$1');
-  return src;
-}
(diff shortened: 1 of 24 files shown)


zertosh:
For a better diff w/o the whitespace changes: https://github.com/mozilla/source-map/pull/202/files?w=1

fitzgen:
I haven't overlooked or forgotten this PR, I just need to malloc a larger block of time to really start digging in.

fitzgen:
Doing #205 instead, which avoids the blame game but should still satisfy all your needs.

zertosh:
Awesome! Thanks! Very clever solution to the indentation problem :smile:

PR #205, opened 2015-08-28
Move off of dryice and to webpack for builds

fitzgen:
We were pretty much the last dryice user, and it isn't worth being on such a
minority build tool when we don't really want to maintain it
ourselves. Additionally, this lets us stop using amdefine and the AMD format,
which was an alright choice in 2011 but has increasingly shown its age. Regular
Common JS modules are used now, which should pave the way for better tree
shaking and unused submodule removal tooling. Note that in order to avoid
ruining `git blame`, this leaves each module indented by two spaces as it was
from the AMD `define` function, and wraps a block around them so that editors
don't freak out too much.

r? @jlongster 

cc @sokra


fitzgen:
Looks like we will need to drop node 0.8.X for webpack which is fine by me. Its years old.

webpack.config.js
```diff
@@ -0,0 +1,85 @@
...
+function run_test() {
+  for (var k in SOURCE_MAP_TEST_MODULE) {
+    if (/^test/.test(k)) {
+      SOURCE_MAP_TEST_MODULE[k](assert);
+    }
+  }
+}
+
+testFiles.forEach(function (file) {
+  module.exports.push({
```
jlongster: You're building each test file with webpack, which includes a `require` of the sourcemap lib. Won't that generate test files that each bundle in the entire sourcemap lib?

If you're generating files that are meant to run on our try server, you probably want to add an alias that resolves the sourcemap library to whatever the path is in firefox.

webpack.config.js
```diff
@@ -0,0 +1,85 @@
...
+    entry: path.join(testsDir, file),
+    output: {
+      path: path.join(distDir, "test"),
+      filename: file.replace(/\-/g, "_"),
+      library: "SOURCE_MAP_TEST_MODULE"
```
jlongster: what does this do? you don't have to have the `library` entry, you can just drop it if this is a dummy value

jlongster:
Personally I'd say to just re-indent everything. You can pass `-w` to `git blame` to just ignore changes that are just whitespace, which you probably want to make the default anyway.

My main concern is building the test files. r+ if you make sure it's not actually bundling in the sourcemap library for all of them, which doesn't seem desirable

webpack.config.js
jlongster: Oh, ignore this comment. I remember what this option does now. It exports the object and names it that.

webpack.config.js
jlongster: Ok, I just looked at the test files and they `require` specific individual modules from the project. So you probably intended to just bundle them all together. Makes sense.

jlongster:
I looked at it further and I understand what you're doing now. r+

fitzgen:
Thanks for the review!

I'd rather not re-indent, even with `git blame -w` for the sake of ease of merging outstanding PRs and branches in the future.

webpack.config.js
fitzgen: Yeah some duplication, but only in tests. The build process is just so much more sane than it used to be, that I think this is worth it.

zertosh:
@fitzgen Can you publish this please?

fitzgen:
I will later today, after @jlongster reviews the follow ups in #206.

jlongster:
@fitzgen You can publish it (I only had 1 or 2 nits I think) but followup with looking into TravisCI for generating builds if you want later, doesn't need to block it.

fitzgen:
Its been published.

PR #213, opened 2015-10-20
Include `dist/` in package.json files

### LINKED ISSUE #212: Include dist in package.json files
I'm working on a [library](https://github.com/mason-lang/esast) that requires source-map. I want this library to work both in node.js and in the browser. This means I want the same code to work with npm and with bower.

For npm, the package.json includes this code:

	"files": [
		"source-map.js",
		"lib/"
	]

So when I run `npm install mozilla/source-map`, I get `lib` but not `dist`.

For bower, I can't use `lib` because that uses CommonJS-style modules and I want to use AMD. So I have to use `dist`.

The problem would be solved if `dist` were included in the package.json `files` so that I could use the same import statement (to `source-map/dist/source-map.min.js`) on the web and in node.

fitzgen:
@andy-hanson, is this enough to fix #212 for you?

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
   "main": "./source-map.js",
   "files": [
     "source-map.js",
-    "lib/"
+    "lib/",
+    "dist/"
   ],
   "engines": {
     "node": ">=0.10.0"


andy-hanson:
Yes, this branch works for me.

fitzgen:
Just published 0.5.2

PR #220, opened 2016-01-07
allGeneratedPositionsFor shouldn't do fuzzy mapping.

ejpbruel:
See the following bug for an explanation of this:
https://bugzilla.mozilla.org/show_bug.cgi?id=1237073

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -208,13 +208,11 @@
         var mapping = this._originalMappings[index];
 
         if (aArgs.column === undefined) {
-          var originalLine = mapping.originalLine;
-
           // Iterate until either we run out of mappings, or we run into
           // a mapping for a different line than the one we found. Since
           // mappings are sorted, this is guaranteed to find all mappings for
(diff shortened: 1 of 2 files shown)


fitzgen:
This is a backwards incompatible change to something that has been in place since day 1 in this library -- any easy way to avoid that so we don't inconvenience the large number of npm dependents? A `fuzzy: bool` property on `aArgs` or another parameter to control the fuzzy mapping?

jlongster:
At some point we're going to have to do some backwards incompatible changes.. otherwise we'll end up with a lot of confusing options. This behavior seems a lot more sane to me, and consistent with the `generatedPositionFor` that returns `null` values if something doesn't map.

I don't really know the extent how this would affect dependents, but my guess it's not very bad and might make things more accurate. Hard to tell though.

ejpbruel:
If backwards compatibility is a concern, I think fitzgen's suggestion is reasonable. Just provide a parameter that will not change the default behavior, but still allows us to have the behavior we need for our use case.

PR #228, opened 2016-04-20
Major performance improvements in serialization

ashaffer:
This PR does two things:

  * Reverts back to #31 to avoid allocating new strings for every key lookup
  * Smarter string concatenation in serializeMappings (avoids re-allocating the big string incrementally).

Together, on my machine at least, this roughly triples the 'serialize source maps' benchmark performance.


fitzgen:
Looks like there are issues on node 0.10:

```
$ npm test

> source-map@0.5.3 test /home/travis/build/mozilla/source-map
> node test/run-tests.js

FAILED ./test-array-set: test that you can add __proto__; see github issue #30!

Error: No element indexed by 0
    at ArraySet_at [as at] (/home/travis/build/mozilla/source-map/lib/array-set.js:88:11)
    at Object.exports.test that you can add __proto__; see github issue #30 (/home/travis/build/mozilla/source-map/test/test-array-set.js:63:28)
    at run (/home/travis/build/mozilla/source-map/test/run-tests.js:21:31)
    at Object.<anonymous> (/home/travis/build/mozilla/source-map/test/run-tests.js:54:12)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)

118 / 119 tests passed.
```

ashaffer:
Alright I pushed a fix for it. Since `__proto__` is the *only* key for which this is an issue, it just checks explicitly for proto and if so transforms it in the same way (accounting for the fact that you could also have the legitimate string '$__proto__', etc..).

This fix hurts the performance slightly, but it's still over twice as fast as before.

lib/util.js
```diff
@@ -235,12 +235,13 @@
    * @param String aStr
    */
   function toSetString(aStr) {
-    return '$' + aStr;
+    if (/^\$*__proto__$/.test(aStr)) return '$' + aStr;
```
fitzgen: Nitpick: match the existing style:

```js
if (condition) {
  then();
}
```

Did you see if pulling the regex definition out of the method helped performance at all? I don't know whether most JS engines would see through this and avoid allocating a `RegExp` object or not (I know spidermonkey will at least re-use and share the compiled regex at a source location between `RegExp` objects, but I don't know if it would avoid allocating the `RegExp` object all together in this particular case).

I suspect it is faster either way to test the string "by hand" rather than with a regular expression:

```js
function isProtoString(s) {
  var length = s.length;

  if (length < 9 /* "__proto__".length */) {
    return false;
  }

  if (s.charCodeAt(length - 1) !== 95  /* '_' */ ||
      s.charCodeAt(length - 2) !== 95  /* '_' */ ||
      s.charCodeAt(length - 3) !== 111 /* 'o' */ ||
      s.charCodeAt(length - 4) !== 116 /* 't' */ ||
      s.charCodeAt(length - 5) !== 111 /* 'o' */ ||
      s.charCodeAt(length - 6) !== 114 /* 'r' */ ||
      s.charCodeAt(length - 7) !== 112 /* 'p' */ ||
      s.charCodeAt(length - 8) !== 95  /* '_' */ ||
      s.charCodeAt(length - 9) !== 95  /* '_' */) {
    return false;
  }

  for (var i = length - 10; i >= 0; i--) {
    if (s.charCodeAt(i) !== 36 /* '$' */) {
      return false;
    }
  }

  return true;
}
```

lib/util.js
```diff
@@ -235,12 +235,13 @@
...
+    return aStr;
   }
   exports.toSetString = toSetString;
-
-  function fromSetString(aStr) {
-    return aStr.substr(1);
+   function fromSetString(aStr) {
+    if (/^\$*__proto__$/.test(aStr)) return aStr.substr(1);
```
fitzgen: Ditto regarding style.

fitzgen:
Thanks! I will merge once the inline comments are addressed.

ashaffer:
Alright, pushed. Fixed the style issues, and you're right, manually checking the string seems to make it about 10% faster or so.

Would it be totally wrong to actually check to see if the JS engine we're in supports the proto thing or not? E.g.

```javascript
function supportsNullProto () {
  var obj = Object.create(null);
  return !('__proto__' in obj)
}

function identity (s) { 
  return s
}

exports.toSetString = supportsNullProto() ? identity : toSetString
exports.fromSetString = supportsNullProto() ? identity : fromSetString
```

fitzgen:
Merged, thank you!

> Would it be totally wrong to actually check to see if the JS engine we're in supports the proto thing or not?

Sure! Feel free to open a follow up PR!

PR #229, opened 2016-04-22
Only prefix proto keys if browser doesn't support null proto

ashaffer:
Hey @fitzgen I made that other change to only prefix/check keys if the JS engine doesn't support a proper null prototype.

Also, would you mind bumping the version and publishing it? I'd like to be able to start using it in browserify.


fitzgen:
Merged as https://github.com/mozilla/source-map/commit/787738914912a1dc047805fbea33c3f3b96444f8 and published to npm. Thanks!

Issue #231, opened 2016-04-22
Possible regression introduced in 0.5.4 release (on Node 0.10)


redonkulus:
We are using webpack 1.x which uses uglify-js under the hood to compile our web application. uglify-js uses the source-map library to generate its source maps. However, it looks like the with the release of version 0.5.4, this caused a regression in Node 0.10 (for some reason not in 0.12). 

Specifically the line in webpack that throws an error:
https://github.com/webpack/webpack/blob/master/lib/optimize/UglifyJsPlugin.js#L66

```
ERROR in main.901a1d3558e2d907db4b.min.js from UglifyJs
TypeError: Cannot read property '1' of null
```

Perhaps the 0.5.4 release changed how the line numbers are returned. 

Is it possible to unpublish this release and re-release it under 0.6.0 to remove the regression?

I'm working on regression test for you to see. But this blocks our CI pipeline currently. Thanks!

redonkulus:
@fitzgen

redonkulus:
When I lock down source-map to 0.5.3 under webpack, the build works correctly. When I pull down this repository locally, running node 0.10 and run the tests, the tests do pass. But maybe that is because this issue isn't covered in the tests, I'm not sure yet.

redonkulus:
Pulling in @ashaffer as well since part of the release was their PR.

redonkulus:
Dependency list:

```
└─┬ webpack@1.13.0
  ├─┬ memory-fs@0.3.0
  │ └─┬ readable-stream@2.1.0
  │   └─┬ unreachable-branch-transform@0.5.1
  │     └─┬ recast@0.11.5
  │       └── source-map@0.5.4
  ├─┬ uglify-js@2.6.2
  │ └── source-map@0.5.4
  ├─┬ watchpack@0.2.9
  │ └─┬ chokidar@1.4.3
  │   └─┬ readdirp@2.0.0
  │     └─┬ readable-stream@2.1.0
  │       └─┬ unreachable-branch-transform@0.5.1
  │         └─┬ recast@0.11.5
  │           └── source-map@0.5.4
  └─┬ webpack-core@0.6.8
    └── source-map@0.4.4
```

ashaffer:
Oh no. Please let me know if you can create a failing test case. I was pretty sure the changes I made shouldn't change it's behavior at all, but its certainly possible there's a weird case I hadn't considered.

ashaffer:
[...]

PR #232, opened 2016-04-23
Fix numerical key bug

ashaffer:
References #231


OleksiiZubko:
works for me

fitzgen:
`ArraySet` only allows string keys and I don't want to change that -- fixing this earlier in the pipeline by stringifying the name given to `addMapping` in https://github.com/mozilla/source-map/pull/235

PR #234, opened 2016-04-25
fix identity is not defined - fixes #233

### LINKED ISSUE #233: crashing on browser
Have no idea how this library got included in my build using browserify, but the following line:

https://github.com/mozilla/source-map/blob/7c4905497b5f0a62fd6a3d3f786ba8eac0db4615/lib/util.js#L258

`ReferenceError: identity is not defined`

Breaks everything now. I don't know what library has this as a dependency

borriglione:


fitzgen:
Merged in 9363979 -- thanks!

PR #235, opened 2016-04-25
Allow string names

### LINKED ISSUE #231: Possible regression introduced in 0.5.4 release (on Node 0.10)

fitzgen:
The names were always supposed to be strings, but because of implementation details numbers just happened to work, and other libraries started relying on this behavior to the point that we have to support it now.

Fixes #231 

r? @ejpbruel


test/test-source-map-generator.js
```diff
@@ -738,4 +738,15 @@
...
+
+  exports['test numeric names #231'] = function (assert) {
+    var generator = new SourceMapGenerator();
+    generator.addMapping({
+      source: 'a.js',
+      generated: { line: 1, column: 10 },
+      original: { line: 1, column: 10 },
+      name: 8
+    });
+    assert.ok(generator.toJSON(), "Adding a mapping with a numeric name did not throw");
```
ejpbruel: Should we also assert here that the numeric name was stringified correctly?

ejpbruel:
Patch looks good to land to me.

fitzgen:
Fixed in b9de0424a6d701f9b10af8c3d8fc0e36d3483344.

OleksiiZubko:
I still have an issue: 
`source-map\lib\util.js:275
  if (s.charCodeAt(length - 1) !== 95  /* '_' */ ||
        ^
TypeError: Object 1223 has no method 'charCodeAt'`, 

because in this case "s" is a Number. If I do in line 265 line:

` if (typeof s !== 'string') { return false } ` - works well

fitzgen:
@OleksiiZubko, can you open a new issue with the full stack trace, minimal testcase, and steps to reproduce? Thanks!

chesleybrown:
We're getting the same issue as you @OleksiiZubko. Here is our full stack trace.

```
TypeError: Object 32 has no method 'charCodeAt'
    at isProtoString (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/node_modules/source-map/lib/util.js:275:9)
    at Object.toSetString (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/node_modules/source-map/lib/util.js:247:7)
    at ArraySet_add [as add] (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/node_modules/source-map/lib/array-set.js:49:19)
    at Function.ArraySet_fromArray [as fromArray] (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/node_modules/source-map/lib/array-set.js:28:9)
    at SourceMapConsumer.BasicSourceMapConsumer (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/node_modules/source-map/lib/source-map-consumer.js:323:26)
    at new SourceMapConsumer (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/node_modules/source-map/lib/source-map-consumer.js:22:7)
    at Concat.add (/var/www/app/node_modules/gulp-concat/node_modules/concat-with-sourcemaps/index.js:56:24)
    at DestroyableTransform.bufferContents [as _transform] (/var/www/app/node_modules/gulp-concat/index.js:71:12)
    at DestroyableTransform.Transform._read (/var/www/app/node_modules/gulp-concat/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:184:10)
    at DestroyableTransform.Transform._write (/var/www/app/node_modules/gulp-concat/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:172:12)
```

Same situation, `s` is an Int for us.

I'm still digging deeper to see if it's something that should be fixed in `concat-with-sourcemaps` or in `source-map`...

chesleybrown:
This issue was fixed in #232, but the fix was rejected. 😢

fitzgen:
Thank you for the stack, fixing in https://github.com/mozilla/source-map/pull/236

OleksiiZubko:
Thanks, guys. Yes, I'm using with concat-with-sourcemaps too.

PR #237, opened 2016-04-27
s/consequtive/consecutive/g - fixing a spelling mistake

kusold:
I recently made the same exact spelling mistake in my code, and a grep through my code base revealed it here too. I didn't run `npm run build`, but can if needed.


fitzgen:
Thanks!

PR #238, opened 2016-06-09
Add source-map typings

roblourens:
TypeScript is now able to find typings that are included with npm packages, using the `typings` property in `package.json`. It would be great if source-map would include typings this way. With this, people using this library with TS will have intellisense and type checks out of the box - much easier than configuring tsd/typings and adjusting their build process.

These typings are based on https://github.com/typed-typings/npm-source-map with some updates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,5 +67,6 @@
   "devDependencies": {
     "doctoc": "^0.15.0",
     "webpack": "^1.12.0"
-  }
+  },
+  "typings": "source-map"
 }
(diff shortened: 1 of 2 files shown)


fitzgen:
I've been busy at an offsite and I'm about to go on vacation. Will look into this when I get back, but I haven't missed it.

jasonLaster:
@ejpbruel I did a quick pass and this looks good.

mootari:
This `package.json` entry is missing the `.d.ts` extension. Isn't the extension mandatory?

roblourens:
I don't think so. And it works for me without.

PR #239, opened 2016-06-14
Fixed "readFileSync" example

bennyn:
The method `fs.readFileSync` returns a Buffer when no encoding is specified (https://nodejs.org/api/fs.html#fs_fs_readfilesync_file_options) and this causes the `SourceMapConsumer` to fail because it expects a string to execute `JSON.parse`. Therefore an encoding needs to be specified, otherwise SourceMapsConsumer fails with `Error: "version" is a required argument`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,7 +566,7 @@ Creates a SourceNode from generated code and a SourceMapConsumer.
   should be relative to.
 
 ```js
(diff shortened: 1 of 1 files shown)


fitzgen:
Thanks!

PR #240, opened 2016-06-16
testing in newer node versions and on osx

MikailBag:


fitzgen:
Awesome, thanks! I didn't know that travis opened up osx machines to open source projects.

PR #244, opened 2016-08-07
Forced installation of latest npm, cached deps

amilajack:


fitzgen:
I'm no Travis CI wizard -- can you explain what these changes do?

amilajack:
1. Cached node_modules for faster tests
2. Forced latest patch version of npm to prevent using buggy versions

tivac:
Why is it doing a global install?

amilajack:
@tivac npm needs to be globally installed.

fitzgen:
Ok, thanks for the explanation!

Issue #250, opened 2016-11-04
SourceMapGenerator.fromSourceMap does not work when mappings is empty


dmurat:
With TypeScript compiler configured to use `system` modules, I'm getting source maps with empty `mappings` when original files are one liners (i.e. `export * from './some/es6/module';`). These source maps have their `sources` and `sourcesContent` collections filled in correctly. Chrome does not have any problems with such source maps, meaning they are used and displayed correctly.

With such source maps, `SourceMapGenerator.fromSourceMap` returns a generator without `_sources` collection set, which I presume is a bug.

To reproduce this behavior, I added following test in this project:

    // in test/test-source-map-generator.js
    exports['test .fromSourceMap with empty mappings'] = function (assert) {
      var map = SourceMapGenerator.fromSourceMap(
        new SourceMapConsumer(util.testMapEmptyMappings));
      util.assertEqualMaps(assert, map.toJSON(), util.testMapEmptyMappings);
    };

    // in test/utils.js
    exports.testMapEmptyMappings = {
      version: 3,
      file: 'min.js',
      names: [],
      sources: ['one.js', 'two.js'],
      sourcesContent: [
        ' ONE.foo = 1;',
        ' TWO.inc = 2;'
      ],
      sourceRoot: '',
      mappings: ''
    };

I can prepare PR with fix (and some more tests) if this is useful.

dmurat:
[Here](https://github.com/mozilla/source-map/pull/251) is a pull request. Tnx.

tromey:
I believe this was fixed by #251

timbell:
I believe there is a related issue [here](https://github.com/mozilla/source-map/blob/ac518d2f21818146f3310557bd51c13d8cff2ba8/lib/source-map-generator.js#L200).

Something like this would fix it I believe:

`var newSources = this._mappings.toArray().length !== 0 ? new ArraySet() : this._sources;`

tromey:
It would be helpful if you could send that change as a PR with a short test case.

timbell:
Ok, will do so over the weekend

PR #251, opened 2016-11-06
Empty mappings fix implemented.

dmurat:
This is a proposed resolution for issue [250](https://github.com/mozilla/source-map/issues/250).

### DIFF
--- a/lib/source-map-generator.js
+++ b/lib/source-map-generator.js
@@ -72,6 +72,17 @@ SourceMapGenerator.fromSourceMap =
       generator.addMapping(newMapping);
     });
     aSourceMapConsumer.sources.forEach(function (sourceFile) {
+      if (aSourceMapConsumer._mappings === '') {
+        var sourceRelative = sourceFile;
+        if (sourceRoot !== null) {
+          sourceRelative = util.relative(sourceRoot, sourceFile);
+        }
+
+        if (!generator._sources.has(sourceRelative)) {
+          generator._sources.add(sourceRelative);
+        }
+      }
+
       var content = aSourceMapConsumer.sourceContentFor(sourceFile);
       if (content != null) {
         generator.setSourceContent(sourceFile, content);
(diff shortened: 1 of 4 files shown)


dmurat:
I updated pull request as I understood your request. I'm hoping that I understood it correctly :-)

lib/source-map-generator.js
```diff
@@ -72,6 +72,17 @@ SourceMapGenerator.fromSourceMap =
       generator.addMapping(newMapping);
     });
     aSourceMapConsumer.sources.forEach(function (sourceFile) {
+      if (aSourceMapConsumer._mappings === '') {
```
ejpbruel: Please remove this if-check. We should always execute the code inside this if-block, even if the mappings are not empty.

test/test-source-map-generator.js
```diff
@@ -252,6 +252,24 @@ exports['test .fromSourceMap with sourcesContent'] = function (assert) {
...
+    new SourceMapConsumer(util.testMapEmptyMappings));
+  util.assertEqualMaps(assert, map.toJSON(), util.testMapEmptyMappings);
+};
+
+exports['test .fromSourceMap with empty mappings and relative sources'] = function (assert) {
+  var map = SourceMapGenerator.fromSourceMap(
+    new SourceMapConsumer(util.testMapEmptyMappingsRelativeSources));
+  util.assertEqualMaps(assert, map.toJSON(), util.testMapEmptyMappingsRelativeSources_generatedExpected);
+};
+
```
ejpbruel: We also need a test for the following scenario:
- We have two sources: foo.js and bar.js
- The mappings are NOT empty, but all the mappings refer only to foo.js.
- The test should assert that bar.js is also included in the list of sources.

Note that this test will fail with your current code, because it will only add bar.js if the mappings are empty.

dmurat:
Ok, I believe that I've understood the issue. I've just updated pull request. Tnx for help :-)

tromey:
I'm sorry about the delay on this.  I think this is fine.

Issue #247, opened 2016-10-18
Cannot read property 'substr' of undefined


kaiza:
TypeError: Cannot read property 'substr' of undefined
    at Function.<anonymous> (...\node_modules\webpack-sources\node_modules\source-map\lib\source-node.js:115:26)

Version: 0.5.6

This problem seems to be related to trying to parse a js file containing a long //# sourceMappingURL at the end. For example, by trying to use the typemoq library in an Angular 2 application, as per this SO issue: 

http://stackoverflow.com/questions/40002699/typemoq-and-angular-cli-dont-work-together#comment67466029_40002699

If I delete the //# sourceMappingURL line from typemoq.js the source node loads fine.

morrisonbrett:
I'm seeing this same error here: https://github.com/mozilla/source-map/blob/master/lib/source-node.js#L95

For me, I can reproduce in an Angular 2 TypeScript file:

THIS breaks:

```
@Component({
  selector: 'tab-bar',
  templateUrl: 'tabs.html'
})
```
THIS works:

`@Component({selector: 'tab-bar',templateUrl: 'tabs.html'})`

jackytse:
Don't write two or more classes in the same ts file. I think it's a bug of 'source-map'.

Aides359:
Manually applying this patch resolves this issue.
http://stackoverflow.com/a/41923610/4068027

edmondtm:
Not only in Ionic-app. I also have the error on my Angular 4 app. When I run ng serve. 

```
92% chunk asset optimization pathtofile/node_modules/source-map/lib/source-node.js:95
          var code = nextLine.substr(0, mapping.generatedColumn -
                             `^`

TypeError: Cannot read property 'substr' of undefined
    at Function.<anonymous> (pathtofile/node_modules/source-map/lib/source-node.js:95:30)
    at Array.forEach (native)
    at BasicSourceMapConsumer.SourceMapConsumer_eachMapping [as eachMapping] (/pathtofile/node_modules/source-map/lib/source-map-consumer.js:155:14)
    at Function.SourceNode_fromStringWithSourceMap [as fromStringWithSourceMap] (/pathtofile/node_modules/source-map/lib/source-node.js:80:24)

```

tromey:
There's a patch you can try - see #257.  Before merging it, though, I'd like a regression test, and I still don't know how to write one.  If you have a recipe for reproducing (the simpler the better, and for best results really walk me through the setup), that would be super helpful.

[...]

PR #257, opened 2017-03-02
fix the bug “TypeError: Cannot read property 'substr' of undefined”

jackytse:
fix the bug “TypeError: Cannot read property 'substr' of undefined”

### DIFF
--- a/lib/source-node.js
+++ b/lib/source-node.js
@@ -91,7 +91,7 @@ SourceNode.fromStringWithSourceMap =
           // There is no new line in between.
           // Associate the code between "lastGeneratedColumn" and
           // "mapping.generatedColumn" with "lastMapping"
-          var nextLine = remainingLines[0];
+          var nextLine = remainingLines[0] || '';
           var code = nextLine.substr(0, mapping.generatedColumn -
                                         lastGeneratedColumn);
           remainingLines[0] = nextLine.substr(mapping.generatedColumn -
@@ -111,7 +111,7 @@ SourceNode.fromStringWithSourceMap =
         lastGeneratedLine++;
       }
       if (lastGeneratedColumn < mapping.generatedColumn) {
-        var nextLine = remainingLines[0];
+        var nextLine = remainingLines[0] || '';
         node.add(nextLine.substr(0, mapping.generatedColumn));
         remainingLines[0] = nextLine.substr(mapping.generatedColumn);
         lastGeneratedColumn = mapping.generatedColumn;


IgorMinar:
could this be merged please?

it fixes quite common issue #247 (also see driftyco/ionic-app-scripts#796 and angular/angular-cli#5476)

in the meantime, one can apply this patch to their project using:

```
(cd node_modules/source-map; curl https://patch-diff.githubusercontent.com/raw/mozilla/source-map/pull/257.patch | patch lib/source-node.js)
```

jasonLaster:
@IgorMinar this looks good to me. Can you include a test that would have failed without the fix?

KeitaElric:
Pls resolve confict and merge it.

jruef:
Hi, could you please resolve the conflicts and merge. I really need this fix!

erykpiast:
Please do!

erykpiast:
Hey @fitzgen, I suppose you're in charge here, so I'll talk to you directly. Sorry for bothering you, but I hope it'll be more effective.

We have an emergency here. Notice that `source-map` isn't direct dependency for the most of people (I guess it's webpack's in many cases) and it's not so easy to just "use forked version until it's fixed properly". Currently I'm modifying my `node_modules` directory three levels deep after installation (even on CI; sounds quite fragile, doesn't it?) and I suppose it's the only way to push their code to production for many people right now.

I've spent quite a lot of time on debugging it to add a test and accelerate the work, without much success though. I did want to share what I've (not) found, but after writing screen-long comment on GitHub my browser crashed and I've lost everything. Shit happens. So the only thing I want to say: please consider to accept this quick, dirty, untested and "I don't really know why it works" fix, as it just resolves serious bug while not breaking any existing test, and try to find proper solution after that. People counts on you. Thank you in advance.

fitzgen:
I don't maintain this repository anymore, that would be @tromey.

lahdekorpi:
Any updates on this?

natejgardner:
Please merge this! Without this fix, webpack is totally broken.

@tromey do you have an ETA?

tromey:
I think this is interesting: https://github.com/mozilla/source-map/issues/247#issuecomment-320233757

michal-filip:
We had the same issue when used as dependency of webpack-sources package. Resolution was to point to this PR explicitly in package.json to override the version webpack-sources uses. Thanks a lot for this!!
`"source-map": "https://github.com/mozilla/source-map/tarball/eb1a675a397cc4752358bec7e84e8e81d7798ac4"`

tromey:
Going to merge this in #288 - same basic patch, just no merge conflict.

Issue #263, opened 2017-05-24
Possible error caused by oversight in source-map-generator.js#L264


Lange:
In [source-map-generator.js#L264](https://github.com/mozilla/source-map/blob/master/lib/source-map-generator.js#L264), a check is made to see if `aOriginal` is falsey. However, I don't think this is ever possible, because `aOriginal` is always an object.

Instead, I think that this should check to see if `aOriginal.line === null && aOriginal.column === null`. When I did this, it seemed to make some pretty nasty errors like https://github.com/Polymer/polymer-bundler/issues/516 go away.

Am I understanding this code correctly? Should I go ahead and open a pull request?

tromey:
I think this can be closed now that #265 was merged.
Please reopen if I'm mistaken about this.

PR #264, opened 2017-05-24
Don't throw when validating a mapping where original.line & original.column are null

Lange:
Hello! :wave:

This addresses #263.

I encountered this issue when feeding sourcemaps from [babili](https://github.com/babel/babili) (babel) into [polymer-bundler](https://github.com/Polymer/polymer-bundler). Another user encountered this issue when just using polymer-bundler on its own (https://github.com/Polymer/polymer-bundler/issues/516).

However, I'm not 100% sure that this pull request is ready to merge just yet. I think this may indicate a problem with other tests. There are several tests that check the case where `aOriginal` is falsey, but I don't think that this is ever possible in real world uses. I think that `aOriginal` will _always_ be an object, but that its `line` and `column` properties might be `null`. In all my tests so far, I have yet to encounter a real-world case that results in a falsey `aOriginal`.

If this is indeed the case, then perhaps some of the existing tests should also be changed?


Lange:
[source-map-consumer.js#L448](https://github.com/mozilla/source-map/blob/master/lib/source-map-consumer.js#L448) looks like the first place where a mapping is defined, and it would indeed be impossible for it to ever be a falsey value. It is always a `Mapping` object.

Lange:
Actually, on further investigation, it might actually be totally possible for `aOriginal` to be falsey in real-world cases. It looks like it's specifically `polymer-bundler` where `aOriginal` is forced to always be a truthy object. Maybe this isn't a bug in `source-map` at all, and is instead a bug in `polymer-bundler`?

https://github.com/Polymer/polymer-bundler/blob/master/src/source-map.ts#L86

Lange:
Ack, sorry, I'm now severely doubting that this is an issue in `source-map`. Doing more research and will follow up in this thread. Right now I think this pull request would break things and introduce bugs.

jasonLaster:
Thanks for the research @Lange. Happy to investigate as well.

jasonLaster:
Thanks for the research @Lange. Happy to investigate as well.

Also, can you remove the dist builds

test/test-source-map-generator.js
```diff
@@ -42,6 +42,13 @@ exports['test adding mappings (case 1)'] = function (assert) {
       generated: { line: 1, column: 1 }
     });
   });
+
+  assert.doesNotThrow(function () {
+    map.addMapping({
+        generated: { line: 1, column: 1 },
+        original: { line: null, column: null }
```
jasonLaster: what is the case where you want to support mapping to a null location?

jasonLaster: My guess is that you have another issue because you ideally want to map to an actual location in an original file.

Lange: Agreed. I'm doing more research now on `polymer-bundler`, which is the tool where my issues are arising from.

Lange:
The root of my issue has been discovered and I've submitted a pull request against the appropriate repo: https://github.com/Polymer/polymer-bundler/pull/519

This PR (and issue #263) _could_ both be closed, but I think this would be a good opportunity to add a helpful error message. This seems like a very easy mistake to make and the resulting error is pretty cryptic. `_validateMapping` could check to see if `aOriginal.line` and `aOriginal.column` are numbers, and throw a more specific error if they aren't.

Or this PR could be merged (after I remove the dist files, as requested) to just allow this usage.

Thoughts?

Lange:
Closing this PR in favor of a separate PR which adds a more informative error message instead.

PR #265, opened 2017-05-24
Add specific error message for when aOriginal is truthy but empty

Lange:
(As discussed in Slack. This PR replaces #264.)

A programmer implementing `source-map` in their own program might think that if they don't specify `original.line` and `original.column` that `source-map` will automatically interpret the mapping as a so-called "Case 1" map, which contains just a generated position and nothing else. However, this is not the case, as the validation in `_validateMapping` is very strict, and expects `aOriginal` to be omitted entirely (or otherwise falsey).

An example of this happening is https://github.com/Polymer/polymer-bundler/pull/519

This PR adds a very specific error message to try to let the person reading the error understand _why_ the error is happening, and what they should probably do to fix it. There's certainly an argument to be made that hyperspecific errors like this don't scale, but seeing as this tripped up people on a project as big as Polymer I figured it might be worth adding.

### DIFF
--- a/lib/source-map-generator.js
+++ b/lib/source-map-generator.js
@@ -259,6 +259,18 @@ SourceMapGenerator.prototype.applySourceMap =
 SourceMapGenerator.prototype._validateMapping =
   function SourceMapGenerator_validateMapping(aGenerated, aOriginal, aSource,
                                               aName) {
+    // When aOriginal is truthy but has empty values for .line and .column,
+    // it is most likely a programmer error. In this case we throw a very
+    // specific error message to try to guide them the right way.
+    // For example: https://github.com/Polymer/polymer-bundler/pull/519
+    if (aOriginal && typeof aOriginal.line !== 'number' && typeof aOriginal.column !== 'number') {
+        throw new Error(
+            'original.line and original.column are not numbers -- you probably meant to omit ' +
+            'the original mapping entirely and only map the generated position. If so, pass ' +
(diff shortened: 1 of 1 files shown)


lib/source-map-generator.js
```diff
@@ -259,6 +259,18 @@ SourceMapGenerator.prototype.applySourceMap =
 SourceMapGenerator.prototype._validateMapping =
   function SourceMapGenerator_validateMapping(aGenerated, aOriginal, aSource,
                                               aName) {
+    // When aOriginal is truthy but has empty values for .line and .column,
+    // it is most likely a programmer error. In this case we throw a very
+    // specific error message to try to guide them the right way.
+    // For example: https://github.com/Polymer/polymer-bundler/pull/519
+    if (aOriginal && typeof aOriginal.line !== 'number' && typeof aOriginal.column !== 'number') {
```
tromey: I wonder if this should check whether either is not a number, like
```
aOriginal && (typeof ... != 'number' || typeof ... != 'number')
```

I wasn't sure.  Let me know what you think; I'll either merge this patch or the obvious follow-up.

tromey:
Also, don't worry about these test failures.

lib/source-map-generator.js
Lange: I had considered that as well, but I think that the case where one (but not both) of the values is a number is a different error, and would warrant a different message.

tromey: Sounds reasonable.

PR #266, opened 2017-05-26
Fix #111 - use incremented index instead of array shift in source-node

### LINKED ISSUE #111: Stop using `shift` in `SourceNode.prototype.fromStringWithSourceMap`
It is O(n) and its inside a loop for each mapping :( very sad. Instead we should just advance an index through the array of remaining lines.

Depends on #103 because that changes the `fromStringWithSourceMap` code, but we will still be using `shift`.

fictitious:
This is only a performance optimization, behavior is unchanged.
This change makes one particular webpack build 2x faster.


fictitious:
Travis failures are for node versions 0.10 and 0.12. I can't find any explanation in the logs. When I ran tests locally with node 0.12.0 npm 2.5.1 and also with node 0.10.32 npm 1.4.28 all tests pass.

PR #268, opened 2017-06-08
undefined error solved

abdulrehmank:
remainingLines[remainingLinesIndex] sometimes give undefined and the build gets failed

### DIFF
--- a/lib/source-node.js
+++ b/lib/source-node.js
@@ -97,7 +97,7 @@ SourceNode.fromStringWithSourceMap =
           // There is no new line in between.
           // Associate the code between "lastGeneratedColumn" and
           // "mapping.generatedColumn" with "lastMapping"
-          var nextLine = remainingLines[remainingLinesIndex];
+          var nextLine = remainingLines[remainingLinesIndex] == undefined ? '' : remainingLines[remainingLinesIndex];
           var code = nextLine.substr(0, mapping.generatedColumn -
                                         lastGeneratedColumn);
           remainingLines[remainingLinesIndex] = nextLine.substr(mapping.generatedColumn -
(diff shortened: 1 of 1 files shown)


tromey:
Thanks for the patch.  This bug has been reported a few times, and there's another PR open for it (since obsoleted by other changes) -- but what is missing is a test case.  I'd like to have one before merging this.  I'm willing to write the test myself but I don't know how to reproduce the bug.  Could you either write a test or explain how I can do so?

fictitious:
From my limited understanding, it happens when source code passed to `fromStringWithSourceMap` does not correspond to the sourcemap in `SourceMapConsumer`.

If you start by adding this test to `test-source-node.js` (which passes)

```
exports['test .fromStringWithSourceMap() empty code'] = function (assert) {
  var emptyCode = '';
  var node = SourceNode.fromStringWithSourceMap(
      emptyCode,
      new SourceMapConsumer(util.emptyMap));

  var result = node.toStringWithSourceMap({
    file: 'min.js'
  });
  var map = result.map;
  var code = result.code;

  assert.equal(code, emptyCode);
  assert.ok(map instanceof SourceMapGenerator, 'map instanceof SourceMapGenerator');
  map = map.toJSON();
  assert.equal(map.version, util.emptyMap.version);
  assert.equal(map.file, util.emptyMap.file);
  assert.equal(map.mappings, util.emptyMap.mappings);
};
```

then change it and make consumer to use `testSourceMap`:

```
  var node = SourceNode.fromStringWithSourceMap(
      emptyCode,
      new SourceMapConsumer(util.testMap));
```

you get

```
FAILED ./test-source-node: test .fromStringWithSourceMap() empty code!
TypeError: Cannot read property 'substr' of undefined
    at Function.<anonymous> (...source-map/lib/source-node.js:121:26)
```

This is unlikely a valid use case, it probably just exposes some other bug somewhere in the code that uses sourcemap, and that code should be fixed regardless of what is done here.

tromey:
Obsoleting in favor of #288 - same thing, slightly different spelling.

PR #272, opened 2017-09-06
Fixing some comparators

iamstolis:
There were two problems with the comparators in `util.js`: a subtraction was used for `string` properties - this does not work as expected (`'a' - 'b'` is `NaN`) and `null` values were not handled correctly by `strcmp`. It was `strcmp(null, 'a') === -1` and `strcmp('a', null) === -1`, i.e., it produced inconsistent comparisons for the sorting (in other words, it claimed that `null` should go before `a` but also `a` should go before `null`).


tromey:
Thanks for the patch.  This looks good to me.

Issue #227, opened 2016-03-15
SourceMapConsumer.prototype.sourceContentFor throws on mappings with `sourceRoot`


Gozala:
Here is a reproducible test case

``` js
var SourceMapConsumer = require('source-map').SourceMapConsumer;
var consumer = new SourceMapConsumer({
  version: 3,
  sources: [ 'index.js' ],
  names: [],
  mappings: ';;AAAA,IAAI,OAAO,MAAP',
  file: 'index.js',
  sourceRoot: './src/',
  sourcesContent: [ 'var name = "Mark"\n' ]
});
consumer.sourceContentFor(consumer.sources[0]);
```

Gozala:
I tried to figure out the issue & it seems to be related to `util.relative` which behaves as follows with a given inputs:

```js
util.relative('./src', 'src/index.js') => 'src/index.js'
```

tromey:
One possible fix is to call `normalize` on the `sourceRoot` that is passed to `sourceMapConsumer`.  I don't know if this is always advisable, but it seems relatively safe.

PR #280, opened 2017-09-22
Issue #227

tromey:


tromey:
Dang it, those patches shouldn't be there.

test/test-source-map-consumer.js
```diff
@@ -1119,3 +1119,16 @@ exports['test consuming names and sources that are numbers'] = function (assert)
   });
   assert.equal(i, 1);
 };
+
+exports['test from issue 227'] = function (assert) {
```
gregtatum: I would appreciate a little more verbosity on the test name. `test non-normalized sourceRoot values (from issue #227)`.

test/test-source-map-consumer.js
```diff
@@ -1119,3 +1119,16 @@ exports['test consuming names and sources that are numbers'] = function (assert)
...
+  var consumer = new SourceMapConsumer({
+    version: 3,
+    sources: [ 'index.js' ],
+    names: [],
+    mappings: ';;AAAA,IAAI,OAAO,MAAP',
+    file: 'index.js',
+    sourceRoot: './src/',
+    sourcesContent: [ 'var name = "Mark"\n' ]
+  });
+  consumer.sourceContentFor(consumer.sources[0]);
```
gregtatum: I'd also like to see an assertion for the new behavior you've introduced here, since the output will not be exactly the same.

```
assert.equal(consumer.sourceRoot, 'src/');
```

PR #282, opened 2017-09-22
Change sourceRoot resolution to match the spec

### LINKED ISSUE #62: Wrong sourceRoot handling?
From the spec:

The spec say that the `sourceRoot` is **prepended**, but mozilla/source-map concat it with an additional slash `/`.

tromey:
The source-map library has long implemented sourceRoot resolution using
URL resolution.  However, the source map specification is reasonably
clear that concatenation is to be used instead.

There is more information in these bugs:
    https://github.com/mozilla/source-map/issues/62
    https://bugzilla.mozilla.org/show_bug.cgi?id=977463

This patch changes the resolution to match what Chrome does.  This is an
incompatible change; but, I believe, better for interoperability.

This patch also adds an optional sourceMapURL parameter to the
SourceMapConsumer constructor.  This allows us to implement the full
source URL resolution algorithm in the source-map library.

Fixes #62


dist/source-map.js
```diff
@@ -1671,6 +1693,10 @@ return /******/ (function(modules) { // webpackBootstrap
 	    throw new Error('Unsupported version: ' + version);
 	  }
 
+	  if (aSourceMapURL) {
+	    sourceRoot = util.join(aSourceMapURL, sourceRoot);
```
tromey: This is a bit iffy given the source map spec.  Perhaps we should delay this computation.

tromey:
This approach is flawed.  It doesn't work if the sourceRoot is empty but the sourceMapURL is not, because it does the operations in an invalid order.

PR #283, opened 2017-09-25
Remove entire first line when JSON XSSI prevention seen

### LINKED ISSUE #278: Strip )]} from JSON before parsing
The source map spec allows prepending ")]}" to avoid XSSI attacks. It says:

> Thus when delivering source maps over HTTP, servers may prepend a line starting with the string “)]}'” to the sourcemap. If the response starts with this string clients must ignore the first line.

It's not clear how important this is to support, but if it is important, it would be nice if `SourceMapConsumer` handled this.

tromey:
The source map specification says that the entire first line should be
removed when the source map string starts with the JSON XSSI preventing
string.
Fixes #278


lib/util.js
```diff
@@ -423,3 +423,21 @@ function compareByGeneratedPositionsInflated(mappingA, mappingB) {
...
 }
 exports.compareByGeneratedPositionsInflated = compareByGeneratedPositionsInflated;
+
+/**
+ * Strip any JSON XSSI avoidance prefix from the string (as documented
+ * in the source maps specification), and then parse the string as
+ * JSON.
+ */
+function parseJSON(str) {
+  if (str.startsWith(")]}'")) {
```
julienw: `startsWith` fails in node 0.10/0.12, sadly: https://travis-ci.org/mozilla/source-map/builds/279505665

julienw: I wouldn't mind to stop supporting these versions but then I think you'll need a major version change.

julienw: I'd reuse the previous regexp to do this as a replacement.

lib/source-map-consumer.js
```diff
@@ -14,7 +14,7 @@ var quickSort = require('./quick-sort').quickSort;
 function SourceMapConsumer(aSourceMap) {
   var sourceMap = aSourceMap;
   if (typeof aSourceMap === 'string') {
-    sourceMap = JSON.parse(aSourceMap.replace(/^\)\]\}'/, ''));
```
julienw: note: if you reuse the same regexp (see below), please note that only the parens needs escaping.

lib/util.js
```diff
@@ -423,3 +423,21 @@ function compareByGeneratedPositionsInflated(mappingA, mappingB) {
...
+    var index = str.indexOf("\n");
+    if (index < 0) {
+      // No line, return nothing.
+      return null;
+    }
+    str = str.substring(index + 1);
```
julienw: Wondering if we should then assert that the source map format revision is > 3. I think we don't need to, but it's your call.

lib/util.js
```diff
@@ -423,3 +423,21 @@ function compareByGeneratedPositionsInflated(mappingA, mappingB) {
...
+ * Strip any JSON XSSI avoidance prefix from the string (as documented
+ * in the source maps specification), and then parse the string as
+ * JSON.
+ */
+function parseJSON(str) {
+  if (str.startsWith(")]}'")) {
+    var index = str.indexOf("\n");
+    if (index < 0) {
+      // No line, return nothing.
+      return null;
```
julienw: I'd rather throw an error, just like JSON.parse throws an error if there is no input.

lib/util.js
```diff
@@ -423,3 +423,21 @@ function compareByGeneratedPositionsInflated(mappingA, mappingB) {
...
+    }
+    str = str.substring(index + 1);
+  }
+  return JSON.parse(str);
+}
+exports.parseJSON = parseJSON;
```
julienw: nitpicking: I don't like the name as it looks like the only thing it does is parsing JSON :) Maybe "parseSourceMapInput" ?

lib/util.js
tromey: I'm planning a major version bump anyhow, due to other pending PRs.

0.10 was still seeing new releases last year; so, ugh, I guess we carry on.

lib/util.js
tromey: I think that's done elsewhere.

lib/source-map-consumer.js
tromey: I did do it this way fwiw.

lib/util.js
```diff
@@ -423,3 +423,13 @@ function compareByGeneratedPositionsInflated(mappingA, mappingB) {
...
 }
 exports.compareByGeneratedPositionsInflated = compareByGeneratedPositionsInflated;
+
+/**
+ * Strip any JSON XSSI avoidance prefix from the string (as documented
+ * in the source maps specification), and then parse the string as
+ * JSON.
+ */
+function parseSourceMapInput(str) {
+  return JSON.parse(str.replace(/^\)]}'(?:[^\n]*?\n)/, ''));
```
julienw: Some simplifications:
* You don't need the grouping :)
* you don't need the ungreedy variant `*?` thanks to your character class

so I think `/^\)]}'[^\n]*\n/` works fine and is simpler :)

tromey: Yes, duh.  Thanks.

PR #284, opened 2017-09-25
Url parsing fix

tromey:


tromey:
Darn it, meant to squash.

PR #285, opened 2017-09-25
Slightly loosen urlRegexp

### LINKED ISSUE #267: the "source" returned by "originalPositionFor" is different with "sources" of sourceMap file
I use "originalPositionFor" to get the original position. 

```js
    console.log(JSON.stringify(sourceMap.sources, null, 4));

    var smc = new SourceMap.SourceMapConsumer(sourceMap);
    var result = smc.originalPositionFor(
    {
        line: parseInt(lineNumber),
        column: parseInt(columnNumber)
    });

    console.log(result.source);
```

I print out the "sources" property of sourceMap and the "source" of "originalPosition" results, but they are different, making it hard for me to locate the exact "sourcesContent" of sourceMap.

sourceMap.sources:
```js
[
    "webpack:///static/js/manifest.b7cf97680f7a50fa150f.js",
    "webpack:///webpack/bootstrap 67e184f9679733298d44"
]
```

result.source:
```js
webpack:/webpack/bootstrap 67e184f9679733298d44
```

result.source misses "//", which is quite strange.

Any suggestion? Thx!

tromey:
urlRegexp was requiring that there be no spaces in the path component of
the URL.  However, this is overly strict; and in particular webpack
creates URLs with an embedded space.
Fixes #267


test/test-source-map-consumer.js
```diff
@@ -1134,3 +1134,18 @@ exports['test non-normalized sourceRoot (from issue #227)'] = function (assert)
...
 };
+
+exports['test webpack URL resolution'] = function (assert) {
+  var map = {
+    "version":3,
+    "sources": ["webpack:///webpack/bootstrap 67e184f9679733298d44"],
+    "names":[],
+    "mappings":"CAAS",
+    "file":"static/js/manifest.b7cf97680f7a50fa150f.js",
+    "sourceRoot":""
```
julienw: nit: no need to quote the keys here :) Also, please add spaces after the colons.

(would make sense to add `prettier` eventually)

lib/util.js
```diff
@@ -26,7 +26,7 @@ function getArg(aArgs, aName, aDefaultValue) {
 }
 exports.getArg = getArg;
 
-var urlRegexp = /^(?:([\w+\-.]+):)?\/\/(?:(\w+:\w+)@)?([\w.-]*)(?::(\d+))?(\S*)$/;
+var urlRegexp = /^(?:([\w+\-.]+):)?\/\/(?:(\w+:\w+)@)?([\w.-]*)(?::(\d+))?(.*)$/;
```
julienw: While reviewing, I realized that in `isAbsolute`, you should use `urlRegexp.test(aPath)` instead of `aPath.match(urlRegexp)` because you don't need the matches, so it should be faster. You can do it in this patch or in another patch if you want.

tromey: I made the change.

tromey: Actually, this change causes many regressions.

tromey: ... because I wrote it incorrectly.

PR #286, opened 2017-09-25
Change sourceRoot resolution to match the spec

### LINKED ISSUE #62: Wrong sourceRoot handling?

tromey:
The source-map library has long implemented sourceRoot resolution using
URL resolution.  However, the source map specification is reasonably
clear that concatenation is to be used instead.

There is more information in these bugs:
    https://github.com/mozilla/source-map/issues/62
    https://bugzilla.mozilla.org/show_bug.cgi?id=977463

This patch changes the resolution to match what Chrome does.  This is an
incompatible change; but, I believe, better for interoperability.

This patch also adds an optional sourceMapURL parameter to the
SourceMapConsumer constructor.  This allows us to implement the full
source URL resolution algorithm in the source-map library.

Fixes #62


tromey:
This allows removing a bunch of code from devtools-source-map as well.

tromey:
Rebased, and avoid `startsWith` and `endsWith`

lib/util.js
```diff
@@ -433,3 +433,38 @@ function parseSourceMapInput(str) {
...
+/**
+ * Compute the URL of a source given the the source root, the source's
+ * URL, and the source map's URL.
+ */
+function computeSourceURL(sourceRoot, sourceURL, sourceMapURL) {
+  sourceURL = sourceURL || '';
+
+  if (sourceRoot) {
+    // This follows what Chrome does.
+    if (sourceRoot[sourceRoot.length - 1] !== '/' && sourceURL[0] !== '/') {
```
gregtatum: I agree that we should do what Chrome does here, the spec doesn't seem to define this area very well :-/

lib/util.js
```diff
@@ -433,3 +433,38 @@ function parseSourceMapInput(str) {
...
+      sourceRoot += '/';
+    }
+    // The source map spec specifies plain prepending here.
+    sourceURL = sourceRoot + sourceURL;
+  }
+
+  if (sourceMapURL) {
```
gregtatum: I would add a few more comments here explaining why you are doing this work, and how it conforms to the spec. It feels a bit mysterious to me.

lib/util.js
```diff
@@ -433,3 +433,38 @@ function parseSourceMapInput(str) {
...
+      var index = parsed.path.lastIndexOf('/');
+      if (index >= 0) {
+        parsed.path = parsed.path.substring(0, index + 1);
+      }
+    }
+    sourceURL = join(urlGenerate(parsed), sourceURL);
+  }
+
+  return normalize(sourceURL);
+}
```
gregtatum: This function drives a lot of the internals, and is dependent on behavior both defined from the spec, and implicitly derived from other's implementations. This code gets hit by our existing tests, but I think it warrants having unit tests just for this function that run through the various types of options that can be done here. I took some time to verify my assumptions in the console of how things worked in order to have more confidence in this review. I think some tests could be fairly quick and terse, but would really help in defining the desired behavior here.

```js
assert.equal(
  computeSourceURL("http://correct.com/", "/a/b.js", "http://wrong.com/source.map"),
  "http://correct.com/a/b.js"
)
```

What do you think?

test/test-source-map-consumer.js
```diff
@@ -888,6 +889,22 @@ exports['test github issue #64'] = function (assert) {
...
 
+exports['test full source content with sourceMapURL'] = function (assert) {
+  var map = new SourceMapConsumer({
+    'version': 3,
+    'file': 'foo.js',
+    'sourceRoot': '',
+    'sources': ['original.js'],
+    'names': [],
+    'mappings': 'AACA',
+    'sourcesContent': ['yellow warbler']
```
gregtatum: Your birds are starting to compete with fitzgen's hip hop references.

lib/util.js
tromey: That's a great idea.

PR #288, opened 2017-09-27
Avoid undefined line in fromStringWithSourceMap

### LINKED ISSUE #247: Cannot read property 'substr' of undefined

tromey:
Avoid calling substr on `undefined` in `fromStringWithSourceMap`.
This can happen when passing in a source map that refers to lines past
the end of the passed-in string -- caller error, but let's not die.
Fixes #247


tromey:
Self-approving because it's the same as a different patch that I would have ok'd had it not had a merge conflict.

PR #289, opened 2017-09-28
M-C fixes

tromey:


tromey:
This fixes a problem pointed out by trying to land the bundle in M-C.

lib/source-map-consumer.js
```diff
@@ -413,9 +443,7 @@ BasicSourceMapConsumer.prototype._version = 3;
  */
 Object.defineProperty(BasicSourceMapConsumer.prototype, 'sources', {
   get: function () {
-    return this._sources.toArray().map(function (s) {
-      return util.computeSourceURL(this.sourceRoot, s, this._sourceMapURL);
-    }, this);
+    return this._absoluteSources.slice(0);
```
julienw: nit: you can remove the "0"

tromey:
> Next time, please include a STR so that I know what I review :)

Yes, sorry about that.  I thought maybe the tests would be clear enough, but that's not really as good as explaining.

The bug is that for some methods, passing in an absolute URL (as found in the `sources`) will not work -- because those methods only recognized "old style" relative URLs.

PR #296, opened 2017-10-27
Updated a comment on quicksort's run-time complexity.

thequicksort:
The lower-bound on deterministic comparison-based sorting algorithms is Ω(n log n) [1], as opposed to O(n). 
[1] - https://www.cs.cmu.edu/~avrim/451f11/lectures/lect0913.pdf

### DIFF
--- a/lib/quick-sort.js
+++ b/lib/quick-sort.js
@@ -68,7 +68,7 @@ function doQuickSort(ary, comparator, p, r) {
     // all the elements that are greater than it after it. The effect is that
     // once partition is done, the pivot is in the exact place it will be when
     // the array is put in sorted order, and it will not need to be moved
-    // again. This runs in O(n) time.
+    // again. This runs in O(n log n) time.
 
     // Always choose a random pivot so that an input array which is reverse
     // sorted does not cause O(n^2) running time.


lib/quick-sort.js
```diff
@@ -68,7 +68,7 @@ function doQuickSort(ary, comparator, p, r) {
     // all the elements that are greater than it after it. The effect is that
     // once partition is done, the pivot is in the exact place it will be when
     // the array is put in sorted order, and it will not need to be moved
-    // again. This runs in O(n) time.
+    // again. This runs in O(n log n) time.
```
tromey: I read the comment just now (I hadn't ever before ...) and I think this refers specifically to the step where the elements are compared to the pivot -- which is O(n).  Though, I had to read this twice, so I think it's a bit confusing, especially because a later comment refers to O(n^2).

I think mentioning O(n log n) makes sense, but not here - maybe in the jsdoc before the function?  And, instead, maybe change this spot to say "This particular step runs in O(n) time"?

fitzgen: This whole function can probably removed now that SpiderMonkey self-hosts Array.prototype.sort w/ custom comparator functions.

Worth re-benchmarking, at least.

fitzgen:
The quick sort is gone; using Array.prototype.sort now.

PR #303, opened 2017-10-31
Fix #302

### LINKED ISSUE #302: SourceNode.fromStringWithSourceMap throws an error on mappings with null sources
Here is an example stacktrace from the exception that is thrown.

```
<omitted>/source-map/lib/util.js:33
  var match = aUrl.match(urlRegexp);
                   ^
TypeError: Cannot call method 'match' of null
    at urlParse (<omitted>/source-map/lib/util.js:33:20)
    at Object.join (<omitted>source-map/lib/util.js:148:18)
    at addMappingWithCode (<omitted>/source-map/lib/source-node.js:155:18)
    at Function.<anonymous> (<omitted>source-map/lib/source-node.js:92:11)
    at Array.forEach (native)
    at SourceMapConsumer_eachMapping [as eachMapping] (<omitted>/source-map/lib/source-map-consumer.js:155:14)
    at Function.SourceNode_fromStringWithSourceMap [as fromStringWithSourceMap] (<omitted>/source-map/lib/source-node.js:86:24)
    at <omitted>...

```

nholland94:
Fixes #302 by adding an extra check for null sources.


lib/source-node.js
```diff
@@ -151,7 +151,7 @@ SourceNode.fromStringWithSourceMap =
       if (mapping === null || mapping.source === undefined) {
         node.add(code);
       } else {
-        var source = aRelativePath
+        var source = aRelativePath && mapping.source
```
tromey: I suspect that in the case where `aRelativePath` is non-null and `mapping.source` is null, that this should evaluate to `aRelativePath`.

So maybe this ought to be fixed in `util.join` instead; that function already tries to handle the empty string, changing it to handle `null` and/or `undefined` as well seems reasonable.

What do you think?

nholland94: That makes sense. I will change that and push up the new change soon.

codehag: This pr has been open for a long time, what is the current status? is it still relevant?

tromey: I think the bug still exists, but this patch still needs an update and a test case.

jasonLaster:
Ping @nholland94 do you think you could help add a test?

insfrag:
I propose
```js
  assert.equal(libUtil.join("a", null), "a");
  assert.equal(libUtil.join("a", undefined), "a");
```
Just below [this line](https://github.com/mozilla/source-map/blob/ddaac7f00a67aefbf184bae16cb90b30ddc84a67/test/test-util.js#L144).

loganfsmyth:
Do we know where the `null` source value is coming from? It's not really clear that anything other than a string should be allowed, so I wondering if fixing it at the source wouldn't be a better solution.

loganfsmyth:
I've landed #371 which removes all this logic, so I'm going to close this. If this is still an issue, let's continue the discussion in #302

nikolay-borzov:
`null` source value might come from webpack generated code. You can use create-react-app to get source map with null source

PR #306, opened 2018-01-08
Use WebAssembly to speed up SourceMapConsumer

fitzgen:
r? @tromey


fitzgen:
This is not ready to merge yet.

paulirish:
This is awesome. 

I just tried it out in https://github.com/danvk/source-map-explorer and my particular usecase **went from 20.5s to 8.2s**.   Nicely done.  cc @danvk

fitzgen:
Expect a detailed blog post in the coming days... ;)

README.md
```diff
@@ -207,8 +229,23 @@ following attributes:
...
+consumer.destroy();
+```
+
+#### SourceMapConsumer.prototype.destroy()
+
+Free this source map consumer's associated wasm data that is manually-managed.
+
 ```js
-var consumer = new sourceMap.SourceMapConsumer(rawSourceMapJsonData);
+consumer.destroy();
```
tromey: Bummer but I guess there's no good way around it.

bench/bench-shell-bindings.js
```diff
@@ -3,21 +3,28 @@ if (typeof load !== "function") {
...
     var src = fs.readFileSync(file, "utf8");
-    vm.runInThisContext(src);
+    vm.runInThisContext(src, { filename: file });
   };
 }
 
 if (typeof print !== "function") {
-  print = console.log.bind(console);
+  print = function (x = "") {
+    console.log(`${x}`);
```
tromey: Not important but I thought `console.log` was self-bound already.

lib/read-wasm.js
```diff
@@ -0,0 +1,40 @@
+if (typeof fetch === "function") {
+  // Web version of reading a wasm file into an array buffer.
+
+  let mappingsWasmUrl = null;
+
+  module.exports = function readWasm() {
+    if (typeof mappingsWasmUrl !== "string") {
+      throw new Error("You must provide the URL of lib/mappings.wasm by calling " +
+                      "SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) " +
```
tromey: Would it be crazy to just have the wasm be a giant string constant in some `.js` file?

lib/source-map-consumer.js
```diff
@@ -951,21 +748,123 @@ function IndexedSourceMapConsumer(aSourceMap, aSourceMapURL) {
...
-      consumer: new SourceMapConsumer(util.getArg(s, 'map'), aSourceMapURL)
-    }
+    const consumer = new SourceMapConsumer(util.getArg(s, 'map'), aSourceMapURL);
+    return consumer.then(consumer => {
+      return {
+        generatedOffset: {
+          // The offset fields are 0-based, but we use 1-based indices when
+          // encoding/decoding from VLQ.
+          generatedLine: offsetLine + 1,
+          generatedColumn: offsetColumn + 1
```
tromey: Not introduced here, but elsewhere it says that columns are 0-based (what I recall as well), so I wonder if this is a latent bug.

lib/source-map-consumer.js
```diff
@@ -1139,20 +1038,21 @@ IndexedSourceMapConsumer.prototype._parseMappings =
...
       for (var j = 0; j < sectionMappings.length; j++) {
         var mapping = sectionMappings[j];
 
-        var source = section.consumer._sources.at(mapping.source);
-        source = util.computeSourceURL(section.consumer.sourceRoot, source, this._sourceMapURL);
+        var source = util.computeSourceURL(section.consumer.sourceRoot, source, this._sourceMapURL);
         this._sources.add(source);
         source = this._sources.indexOf(source);
 
         var name = null;
```
tromey: I didn't examine more context but I suspect this line should be removed.

lib/source-map-consumer.js
```diff
@@ -1180,4 +1080,156 @@ IndexedSourceMapConsumer.prototype._parseMappings =
...
+      mappings = this._originalMappings;
+      break;
+    default:
+      throw new Error("Unknown order of iteration.");
+    }
+
+    var sourceRoot = this.sourceRoot;
+    mappings.map(function (mapping) {
+      var source = null;
+      if(mapping.source !== null) {
```
tromey: Trivia: space after the "if".

test/test-source-map-generator.js
```diff
@@ -241,42 +241,42 @@ exports['test that source content can be set'] = function (assert) {
   assert.equal(map.sourcesContent[1], null);
 };
 
-exports['test .fromSourceMap'] = function (assert) {
-  var map = SourceMapGenerator.fromSourceMap(new SourceMapConsumer(util.testMap));
+exports['test .fromSourceMap'] = async function (assert) {
+  var map = SourceMapGenerator.fromSourceMap(await new SourceMapConsumer(util.testMap));
```
tromey: I suppose it's not a big deal to leak the consumers in the tests.

lib/source-map-consumer.js
fitzgen: Perhaps? I had to do this to get the tests passing with the wasm. I think we can investigate further in a follow up.

test/test-source-map-generator.js
fitzgen: Hm, I thought I caught them all, did I miss some?

README.md
fitzgen: I have ideas to improve this in follow ups:

Either add an async RAII-ish function like:

```js
SourceMapConsumer.with = async function (rawSourceMap, f) {
  const consumer = await new SourceMapConsumer(rawSourceMap);
  try {
    await f(consumer);
  } finally {
    consumer.destroy();
  }
};
```

Or alternatively give every `SourceMapConsumer` its own wasm module instance, make the `Mappings` a global/implicit in the wasm heap, and let the GC manage lifetimes. Requires further investigation.

bench/bench-shell-bindings.js
fitzgen: I know it differed across engines at one point in time, unsure if it still does or not.

In fact, the shell version of the benchmarks doesn't work anymore (added too many options that are only exposed in the Webpage), so I should just remove it.

lib/read-wasm.js
fitzgen: It would bloat the code size, which I took quite a bit of effort in keeping small.

fitzgen:
> I do wonder if it's possible to remove the need for a separate wasm URL and thus (IIUC) the need for the constructor to return a promise.

Compiling wasm is still done off-thread, and returns a promise, so even if we already had the blob, the constructor would still return a promise.

lib/source-map-consumer.js
fitzgen: The `var name = null;` line? The `name` variable is assigned to and used a little bit down this method.

tromey: Thanks.

test/test-source-map-generator.js
tromey: I didn't see where these consumers were destroyed.  But maybe `fromSourceMap` does and I missed it.

tromey:
> Compiling wasm is still done off-thread, and returns a promise, so even if we already had the blob, the constructor would still return a promise.

Thanks.  Never mind then.

fitzgen:
Ok, I think this is just about ready to land. Waiting on OSX jobs in Travis CI.

fitzgen:
FYI, https://hacks.mozilla.org/2018/01/oxidizing-source-maps-with-rust-and-webassembly/ is live with details on the experience.

Going to go ahead and merge this PR!

CONTRIBUTING.md
```diff
@@ -122,3 +112,51 @@ can use as well:
...
+
+Build Rust crate as a `.wasm` file:
+
+```
+$ cd source-map-mappings-wasm-api/
+$ cargo build --release --target wasm32-unknown-unknown
+```
+
+The resulting `wasm` file will be located at
+`source-map-mappings-c-api/target/wasm32-unknown-unknown/release/source_map_mappings.wasm`.
```
Kagami: s/c-api/wasm-api/

bench/README.md
```diff
@@ -1,15 +1,20 @@
...
 Ensure that you have built the library, as these benchmarks rely on
 `dist/source-map.js`. See the main README.md for detais on building.
 
-## Running Within a Browser
+Run a local webserver from the root of the repository:
 
-Open `bench.html` in a browser and click on the appropriate button.
+```
+$ cd source-map/
+$ python -m SimpleHTTPServer
```
Kagami: Better `python2 -m SimpleHTTPServer` or `python3 -m http.server`

lib/read-wasm.js
Kagami: `initialize()` accepts just string though?

fitzgen: The public `SourceMapConsumer.initialize` function takes an object, the interal `initialize` function in `lib/read-wasm.js` just takes a string.

CONTRIBUTING.md
fitzgen: Nice catch!

bench/README.md
fitzgen: Thanks

PR #308, opened 2018-01-18
Large speed ups in iterating over all mappings

fitzgen:
r? @tromey 

This is beneficial for both the extant JS and the incoming wasm in #306.

Order of magnitude speed up iterating over all mappings by throwing a dumb LRU cache in front of the `normalize` function.

Ignore last commit in review -- all mechanical.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -88,7 +88,25 @@ function normalize(aPath) {
   }
   var isAbsolute = exports.isAbsolute(path);
 
-  var parts = path.split(/\/+/);
+  // Split the path into parts between `/` characters. This is much faster than
+  // using `.split(/\/+/g)`.
+  var parts = [];
+  var start = 0;
+  var i = 0;
+  while (true) {
+    start = i;
+    i = path.indexOf("/", start);
+    if (i === -1) {
+      parts.push(path.slice(start));
+      break;
+    } else {
+      parts.push(path.slice(start, i));
+      while (i < path.length && path[i] === "/") {
+        i++;
+      }
+    }
+  }
+
   for (var part, up = 0, i = parts.length - 1; i >= 0; i--) {
     part = parts[i];
     if (part === '.') {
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -89,6 +89,25 @@ function normalize(aPath) {
   var isAbsolute = exports.isAbsolute(path);
 
   var parts = path.split(/\/+/);
```
bnjbvr: nit: the call to `path.split()` hasn't been removed yet ;)

lib/util.js
```diff
@@ -66,6 +66,43 @@ function urlGenerate(aParsedUrl) {
...
+ * `f`.
+ *
+ * We keep at most `MAX_CACHED_INPUTS` memoized results of `f` alive. The
+ * memoization is a dumb-simple, linear least-recently-used cache.
+ */
+function lruMemoize(f) {
+  const cache = [];
+
+  return function (input) {
+    for (var i = 0; i < cache.length; i++) {
```
tromey: Is this faster than `Array.indexOf`?

tromey:
I couldn't load the "files changed" view for some reason, but this is r+ from me, with the comment in the first patch addressed.  The thing about `Array.indexOf` is optional.  Thanks!

lib/util.js
fitzgen: D'oh

lib/util.js
fitzgen: We can't use `indexOf` here because the data in the array is an `Object` and so will be tested for pointer equality.

We could use `Array.prototype.find`, but now we're introducing functions and things, and the whole point of this code is to gain whatever control we can over execution, which higher-order-functions can be a bit anathema to.

fitzgen:
Thanks for review!

fitzgen:
The queue for OSX machines in Travis is very long, going to go ahead and merge this now, since Linux is OK and it WFM on OSX locally.

PR #309, opened 2018-01-19
Get ready for new release on npm

fitzgen:
I think this is all that's needed, and after this merges we can publish!

r? @tromey


fitzgen:
OSX backlog on travis ci is still super long; merging now.

fitzgen:
Thanks for review!

pkkid:
Just a heads up. This change breaks backwards compatibility in a minor version release. Most people (as well as the default settings in package.json files) expect breaking changes to have a major version change. We figured it out by forcing the old version to be installed 0.6.0. There may be other people coming in here with similar issues.

pkkid:
That makes sense, sorry for the noise.

PR #312, opened 2018-01-20
🐒 Fix use on the web documentation to actually work

samccone:
There are two fixes:

The old path is no longer part of the dist folder.
The new package that uses wasm requires the user to initialize the path to the lib for things to work.


samccone:
Thank you!

PR #321, opened 2018-02-13
Update source-map TypeScript typings to 0.7.0

jvilk:
First off: Congrats on the excellent 0.7.0 release! I look forward to measuring the performance impact on my memory leak debugger, [BLeak](https://github.com/plasma-umass/bleak).

The TypeScript typings contained within this repository are incorrect for the current version posted to NPM, which has breaking changes from 0.5.0. As a result, TypeScript code is unable to properly interoperate with the latest version of the library.

I have updated the typings to make the SourceMapConsumer constructors return a Promise and to add typings for the new `destroy()` and `with` methods.

I have also added a `types` field to `package.json`, so users of `source-map` can import from this module without requiring the `@types/source-map` metapackage.

**If you accept and merge this pull request, please publish a new version of source-map to NPM with these changes so TypeScript code can interoperate with 0.7.0!** Otherwise, TypeScript code will continue to use the invalid typings file published with the NPM module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "url": "http://github.com/mozilla/source-map.git"
   },
   "main": "./source-map.js",
+  "types": "./source-map.d.ts",
   "files": [
     "source-map.js",
     "source-map.d.ts",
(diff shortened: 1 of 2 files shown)


alexeagle:
Is it actually an advantage that the typings are included in the source-map package now? Unless you add some tests that the typings are correct (we have an example in Angular if you want) - then it's just an unmaintained extra file, and we would be better off having `@types/source-map` maintained by the TypeScript community...

Issue #319, opened 2018-01-31
TypedError in wasm.js


c-niederer:
**ISSUE**
When executing a TypeError occurs in _wasm.js_ because a callback is null.

**EXPECTED**
A defined callback.

**STEPS TO REPRODUCE**
- run `npm install`, `npm run build` and `npm run exec`

This issue came up after porting the code to source-map@0.7.0.

Sample code can be found [here](https://github.com/c-niederer/source-map-bug).

fitzgen:
Thanks for filing an issue!

fitzgen:
Fixed in https://github.com/mozilla/source-map/pull/330

PR #330, opened 2018-04-21
Allow nested consumer usage

ds300:
Hi! :wave:

I was excited to test out the latest source-map for [react-native-typescript-transformer](https://github.com/ds300/react-native-typescript-transformer) and ran into an issue with the way the wasm consumer bindings are written. They assume that the callback passed to, e.g. eachMapping will not make use of another consumer. This resulted in an error which I created a regression test for in this PR, along with a fix.

Cheers!

### DIFF
--- a/dist/source-map.js
+++ b/dist/source-map.js
@@ -2919,7 +2919,7 @@ module.exports = function wasm() {
     return cachedWasm;
   }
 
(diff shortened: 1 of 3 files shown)


ds300:
Looks like this fixes #319 too

PR #332, opened 2018-05-04
Apply the Firefox eslint rules, minus the Firefox-specific plugins.

hildjj:
Fix up the myriad changes.  Justification for the large changeset for quotes: they were inconsistent before.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,327 @@
+"use strict";
+
+module.exports = {
+  "env": {
+    "node": true,
+    "es6": true
+  },
+
+  "parserOptions": {
+    "ecmaVersion": 9
+  },
+
+  "globals": {
+    "fetch": false,
+    "WebAssembly": false
+  },
+
+  "rules": {
+    // Require spacing around =>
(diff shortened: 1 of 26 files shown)


fitzgen:
Also, is there a way to enforce using `const` for variables that are not reassigned? If so, we should apply that.

tromey:
> Also, is there a way to enforce using const for variables that are not reassigned? If so, we should apply that.

This was recently agreed to in devtools: https://github.com/devtools-html/rfcs/issues/44.
The rule is `prefer-const: "error"`.

hildjj:
Note that `npm run lint` now gets called in `npm test`, so travis should pick this up going forward.  See https://travis-ci.org/hildjj/source-map/builds/375031782 for evidence.  I've updated the patch with const/let changes as requested, and added `"no-var": "error"` as well, to make `prefer-const` carry more weight.

Any other eslint rules to add?  May as well put them all into an omnibus patch. :)

fitzgen:
LGTM -- do you have anything else @tromey?

Thanks for sticking with this @hildjj!

PR #336, opened 2018-05-08
First stab at classes

### LINKED ISSUE #259: array-set has terrible memory usage
We have a large project that is build using webpack, which uses source-map to generate the sourcemaps.
There are issues with performance of the phase that generates the sourcemaps so we did some profiling. One thing we noticed right off the bat is the sharp sawtooth shape of the memory usage graph of the node process. This basically means that a lot of the time is spent garbage collecting. Digging deeper, we profiled using the google chrome dev tools and noticed that one of the things allocating a lot of memory is array-set.  It was using gobs of memory by making large source strings keys into an object which causes string copies. We monkey patched it on our side, but it probably makes sense to change this class to use some modern javascript Set/Map data structure so everybody gets the benefit.
Here's the monkey patch we use:
```js
var ArraySet = require("webpack-core/node_modules/source-map/lib/source-map/array-set").ArraySet;

/**
 * the backing map
 * use a map to check uniqueness instead of an object's keys
 * @return {Map}
 */
ArraySet.prototype._map = function ArraySet__map() {
    if (!this._theMap) {
        this._theMap = new Map();
    }

    return this._theMap;
};

/**
 * Return how many unique items are in this ArraySet. If duplicates have been
 * added, than those do not count towards the size.
 *
 * @returns Number
 */
ArraySet.prototype.size = function ArraySet_size() {
    return this._map().size;
};

/**
 * Add the given string to this set.
 *
 * @param String aStr
 */
ArraySet.prototype.add = function ArraySet_add(aStr, aAllowDuplicates) {
    var isDuplicate = this.has(aStr);
    var idx = this._array.length;
    if (!isDuplicate || aAllowDuplicates) {
        this._array.push(aStr);
    }
    if (!isDuplicate) {
        this._map().set(aStr, idx);
    }
};

/**
 * Is the given string a member of this set?
 *
 * @param String aStr
 */
ArraySet.prototype.has = function ArraySet_has(aStr) {
    return this._map().has(aStr);
};

/**
 * What is the index of the given string in the array?
 *
 * @param String aStr
 */
ArraySet.prototype.indexOf = function ArraySet_indexOf(aStr) {
    var idx = this._map().get(aStr);
    if (idx >= 0) {
        return idx;
    }
    throw new Error('"' + aStr + '" is not in the set.');
};

```

### LINKED ISSUE #334: Switch to ES6 classes
In the theme of #332, switch to ES6 classes instead of the older `.prototype` style, for better readability and consistency.  Ensure that documentation is still generated correctly.

If you agree this is something you wouldn't mind taking, assign it to me, and I'll send a PR.

hildjj:
`source-map-consumer.js` is a mess in order to keep the API the same.  That file needs careful review.  Everything else is pretty straightforward.  Fixes #334.  Fixes #259.

Don't merge yet; I want to double-check the coverage and think about it a bit more.  But I'd love a few extra eyeballs on this in the meantime.  Feel free to be brutal in comments.

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -17,7 +17,8 @@ const INTERNAL = Symbol("smcInternal");
 class SourceMapConsumer {
   constructor(aSourceMap, aSourceMapURL) {
     // If the constructor was called by super(), just return Promise<this>.
-    // Yes, this is a hack to retain the pre-existing API.
+    // Yes, this is a hack to retain the pre-existing API of the base-class
+    // constructor also being an async factory function.
     if (aSourceMap == INTERNAL) {
       return Promise.resolve(this);
     }
@@ -1242,11 +1243,10 @@ function _factory(aSourceMap, aSourceMapURL) {
     sourceMap = util.parseSourceMapInput(aSourceMap);
   }
 
-  return Promise.resolve().then(_ => {
-    return sourceMap.sections != null
+  const consumer = sourceMap.sections != null
       ? new IndexedSourceMapConsumer(sourceMap, aSourceMapURL)
       : new BasicSourceMapConsumer(sourceMap, aSourceMapURL);
-  });
+  return Promise.resolve(consumer);
 }
 
 function _factoryBSM(aSourceMap, aSourceMapURL) {
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -12,156 +12,152 @@ const base64VLQ = require("./base64-vlq"); // eslint-disable-line no-unused-vars
 const readWasm = require("../lib/read-wasm");
 const wasm = require("./wasm");
 
-function SourceMapConsumer(aSourceMap, aSourceMapURL) {
-  let sourceMap = aSourceMap;
-  if (typeof aSourceMap === "string") {
-    sourceMap = util.parseSourceMapInput(aSourceMap);
+const INTERNAL = Symbol("smcInternal");
+
+class SourceMapConsumer {
+  constructor(aSourceMap, aSourceMapURL) {
+    // If the constructor was called by super(), just return Promise<this>.
+    // Yes, this is a hack to retain the pre-existing API.
+    if (aSourceMap == INTERNAL) {
+      return Promise.resolve(this);
+    }
+
+    return _factory(aSourceMap, aSourceMapURL);
   }
 
-  return Promise.resolve().then(_ => {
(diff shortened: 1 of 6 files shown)


lib/source-map-consumer.js
```diff
@@ -12,157 +12,153 @@ const base64VLQ = require("./base64-vlq"); // eslint-disable-line no-unused-vars
...
   }
 
-  return Promise.resolve().then(_ => {
-    return sourceMap.sections != null
-      ? new IndexedSourceMapConsumer(sourceMap, aSourceMapURL)
-      : new BasicSourceMapConsumer(sourceMap, aSourceMapURL);
-  });
-}
+  static initialize(opts) {
+    readWasm.initialize(opts["lib/mappings.wasm"]);
```
hildjj: This line isn't covered in the tests.  Does running the tests in the browser depend on #158 ?

lib/source-map-consumer.js
```diff
@@ -199,146 +195,134 @@ exports.SourceMapConsumer = SourceMapConsumer;
...
-  if (version != this._version) {
-    throw new Error("Unsupported version: " + version);
-  }
-
-  if (sourceRoot) {
-    sourceRoot = util.normalize(sourceRoot);
-  }
+class BasicSourceMapConsumer extends SourceMapConsumer {
+  constructor(aSourceMap, aSourceMapURL) {
+    return super(INTERNAL).then(that => {
```
hildjj: You thought `that` was a holdover from before arrow functions.  But in this case, the constructor returns a Promise, which is pretty non-standard.  So hence the workaround.

lib/source-map-consumer.js
```diff
@@ -199,146 +195,134 @@ exports.SourceMapConsumer = SourceMapConsumer;
...
-  function SourceMapConsumer_fromSourceMap(aSourceMap, aSourceMapURL) {
-    return new BasicSourceMapConsumer(aSourceMap.toString());
-  };
+    return -1;
+  }
 
-/**
- * The version of the source mapping spec that we are consuming.
- */
-BasicSourceMapConsumer.prototype._version = 3;
```
hildjj: Used the one off the base class prototype.  Don't know if that's strictly what was desired, but this whole conglomeration of classes is unlikely to work for another version number than 3.

hildjj:
@fitzgen: ready to merge, as far as I can tell.

fitzgen:
Thanks!

PR #339, opened 2018-05-22
 Add option to initialize mappings wasm via ArrayBuffer

rtsao:
This PR adds the ability to directly pass in the mappings wasm into `SourceMapConsumer.initialize`. This is needed for browser usage without needing to make a request (e.g. needing to work around CSP restrictions).

**Example**

```js
const contents = /* the ArrayBuffer contents of mappings.wasm */

sourceMap.SourceMapConsumer.initialize({
  "lib/mappings.wasm": contents
});
```

I didn't implement this because it would be a breaking change, but I think a better alternative API might be something like:

```js
const wasmUrl = "https://unpkg.com/source-map@0.7.3/lib/mappings.wasm"; 

sourceMap.SourceMapConsumer.initialize({
  "lib/mappings.wasm": fetch(wasmUrl).then(res => res.arrayBuffer())
});
```


fitzgen:
Bleh, I overlooked that. We would need some kind of headless browser testing setup. If you want to dig into adding that, please be my guest :)

In the meantime, we can just merge this. Thanks again!

rtsao:
Just did, thanks! 😄

Issue #349, opened 2018-06-24
if (typeof fetch === "function") misidentifies node environment


sprockow:
I am currently working building a node endpoint that accepts a minified stack trace from a client, and then uses this library to find the originating line in the un-minified source using a shared copy of the sourcemap.

The problem I'm running into is that the library uses "fetch" feature detection to determine whether the current environment is a browser or node environment.

`if (typeof fetch === "function") {`

and then conditionally exports a module in the read-wasm file: https://github.com/mozilla/source-map/blob/master/lib/read-wasm.js#L1

The node environment I'm developing within has been configured to use a isomorphic fetch polyfill, so this assumption has caused my node environment to incorrectly use the browser behavior and immediately throw the error: 

> `You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer`

I understand that my environment might be a bit different than most, but I'd like to challenge whether making assumptions like this is the correct way of browser vs. node detection. I came across a clever 
solution described within this stackoverflow question, https://stackoverflow.com/questions/17575790/environment-detection-node-js-or-browser , where you can look up the default top-level function scope of the environment (`window` vs `global`). By looking at the default top level scope, this solution will even handle the case where `window` has been mocked out by a library like jsdom.

I've implemented this in a fork, and it seems to have fixed my problem. Wondering whether this is something you would be open to considering.

Also, if you otherwise think I've done something else completely wrong, please let me know. :)

tromey:
> I've implemented this in a fork, and it seems to have fixed my problem. Wondering whether this is something you would be open to considering.

It seems reasonable to me.

sprockow:
Ok, let me put together a pull request and I'll send this over for review.

sprockow:
Finally go around to doing this. Submitted #350 as a possible solution to this problem.

tromey:
Fixed by #350.

theKashey:
This issue is not solved or closed while fixed only in the beta version.

loganfsmyth:
@theKashey I don't know when we'll be doing further releases, but there should be no problem using the beta release if it's a feature that is important for your usecase.

theKashey:
Unfortunately some people (or their security department) could not afford `beta` releases. But if beta.0 is actually stable - could you publish it as, for example, `next` tag?

karlhorky:
[...]

PR #350, opened 2018-07-11
using top level context to determine node vs browser environment

sprockow:
Possible solution for:
#349 

Ran into this issue because of my node environment's fetch polyfill. To get around this, I used a solution inspired by the solution proposed in this stack overflow question to fix this for myself.


sprockow:
So, this environment detection function is kind of hard to test. It's by definition very tightly coupled with the environment in runs on. It's written to avoid being mocked (so that a  node environment with a mocked window global, like with jsdom, doesn't trigger a false positive browser detection).

I'm open to an argument as to how/why to test this function.

tromey:
This looks reasonable to me, but it occurred to me that I didn't know whether it would work properly in the weird environment that the devtools run in.  So, I think someone else ought to review it, like maybe @codehag

codehag:
This looks good to me 👍

loganfsmyth:
Another option here would be to use a replacement file for browser builds, e.g.

in `package.json`
```
"browser": {
  "lib/read-wasm.js": "lib/read-wasm-browser.js"
}
```
then the alternate file would be used when the library is bundled with Webpack and such, and it would be up to the consumers of the package to choose the build target.

sprockow:
@loganfsmyth - this seems like a much cleaner approach. I haven't done too much with npm publishing, so I'm unfamiliar with this environment specific configuration. Is there a good example/best practice of how webpack projects that target these sorts alternative builds?

loganfsmyth:
Webpack has its [`target` option](https://webpack.js.org/configuration/target/) and that trickles back to the default value for [`resolve.mainFields`](https://webpack.js.org/configuration/resolve/#resolve-mainfields). Since Webpack defaults to `"web"` as it's target, the default behavior is to use the `browser` field to swap out files for alternatives.

I think it'd be a good approach to handle to existing way that this code works. That said, I do want to clarify that I have reservations in general about how this all works, since I think it is a little scary for a library to behave in two very different ways depending on how it is used, since a you have to pass the worker URL in one case but not the other. It might be nicer in the long run for the node build to use `fs` and the browser build to load a JS file containing the Base64-encoded WASM instead of having to use `fetch`, with `fetch` usage being a general opt-in behavior if people do want it still.

PR #354, opened 2018-08-31
Inline wasm module into the browser build

corhere:
Make the existence of the wasm module as much of an implementation
detail in the browser as it is in Node.js by inlining its contents into
the browser build. It is no longer necessary to call
`SourceMapConsumer.initialize()` when using it in a browser environment.


fitzgen:
Thanks for the PR, but this isn't the right solution.

PR #356, opened 2018-09-04
Fixes #355: add absolute path checking to util.join();

### LINKED ISSUE #355: util.join() doesn't handle absolute path on Windows correctly
We use the Webpack pipeline on Windows with sass-loader, postcss-loader, and css-loader. Then we found the generated source paths went crazy: `F:\Projects\web-project-boilerplate\src\scss\index.scss` just becomes `F:/Projects/web-project-boilerplate/src/scss/F:/Projects/web-project-boilerplate/src/scss/F:/Projects/web-project-boilerplate/src/scss/index.scss`.

As I followed the codebase in the pipeline, I noticed that the sass-loader use relative path, then the postcss-loader and css-loader turn to use absolute path for "sources" field in the sourceMap. In addition, the absolute path on Windows starts with a symbol of the disk driver, e.g. `F:` in our case. When it comes to sourceMap's `util.join()`, it only detects absolute path starts with `/`

https://github.com/mozilla/source-map/blob/c3a8f5a2a4cc7ac5bc6c06bc93745d3c57bd8860/lib/util.js#L228

So it joins multiple paths across the loaders.

Also, the `util.isAbsolute` doesn't deal with the Windows absolute path as well.

https://github.com/mozilla/source-map/blob/c3a8f5a2a4cc7ac5bc6c06bc93745d3c57bd8860/lib/util.js#L241

Appendix:
Here is a case of how expressjs test the absolute path:
```
exports.isAbsolute = function(path){
  if ('/' === path[0]) return true;
  if (':' === path[1] && ('\\' === path[2] || '/' === path[2])) return true; // Windows device path
  if ('\\\\' === path.substring(0, 2)) return true; // Microsoft Azure absolute path
};
```

I'm not sure we need to handle the azure absolute path, but I would be better if we had window device path included.

Thanks

mutoo:


mutoo:
@tromey Yeah, I agree with you! The sources should be URLs since they're used by the browser. In our pipeline, the sources are sitting in the local filesystem, so that it'd prefix with file: protocol. Maybe the webpack-internal:: protocol is better.

mutoo:
I removed the test for windows device path in util.isAbsolute() function.
But the util.join() still needs the isAbsolute() to test the absolute path, so I'll keep it.

loganfsmyth:
I've landed #371 which may resolve this. Would you be able to check on master? It's not clear to me what this patch is looking to resolve.

mutoo:
@loganfsmyth I've reviewed the code on master, the bug was fixed.
The issue I tried to fix was that the old `join` function didn't check an absolute path correctly.
This can be close.

Issue #275, opened 2017-09-22
Use built-in URL


tromey:
source-map implements its own URL parsing, but it seems to me that it would be better to just use the built-in `URL`.

tromey:
For node we need `require("url")` but for the browser we don't want that.

tromey:
But maybe we could work around that in devtools-source-map.

tromey:
We'll also need to work around it for the bundle that's put in `devtools/shared/sourcemap`.  Maybe we ought to make a new webpack config for that.

tromey:
One problem the current approach causes is that library users can't pass a `data:` URL as the source map URL to `SourceMapConsumer`, because the current code can't parse a `data:` URL.  (Though, maybe this is minor, because passing a `data:` URL there is useless anyway...)

tromey:
This problem came up again here: https://bugzilla.mozilla.org/show_bug.cgi?id=1451274

I wrote a quick patch to change this but I am not completely sure it is correct.  E.g., it regresses this test:
```
  assert.equal(libUtil.join('http://foo.org/a//', '//b'), 'http://b');
```
and this one:
```
  assert.equal(libUtil.join('http://www.example.com', '//foo.org/bar'), 'http://foo.org/bar');
```
I can't recall offhand if these are part of the spec or not, will have to check.

edmorley:
Hi! Is there anything we can do to help move this forwards?

Several Mozilla web teams use webpack based workflows, so are potentially affected by sourcemap issues that come up (eg webpack-contrib/style-loader#303). It would be good to try and keep our own employees using Firefox for web development :-)

tromey:
> Hi! Is there anything we can do to help move this forwards?

Sure - take over the patch here: https://github.com/tromey/source-map/tree/use-real-url
The main thing is to investigate the issues pointed out above (https://github.com/mozilla/source-map/issues/275#issuecomment-378696129).

For the webpack-contrib issue you linked to, a further part would be to add tests for blob URLs.

If you really want to do all this I can also tell you how to get everything landed.  I'm not working on devtools any more, which is why I'm not doing this... but another option would be to contact someone in devtools to see if anyone there can do it.

jasonLaster:
This bug is a priority.

If we can land this branch https://github.com/mozilla/source-map/compare/master...tromey:use-real-url, here I'd be happy to cut a  new release and land it in MC.

[...]

PR #360, opened 2018-09-28
Use standard-conforming URL

redeyes2015:
Initial work is done by @tromey.

A commit is added so that when `URL` is presented in global object, it is used as
the consturctor; otherwise fallback to use `require('url').URL` in order to support
the environments of node v8 and v9. Since node v10, `URL` is also presented in
global object.


redeyes2015:
The issues metioned in https://github.com/mozilla/source-map/issues/275#issuecomment-378696129 are not checked yet.

By the way, I made another [branch](https://github.com/redeyes2015/source-map/tree/0.6.1-use-real-url) based on 0.6.1 to test in debugger.html, which make one test case fail: devtools-source-map/src/tests/source-map.js line 29: expect `"webpack:///webpack/bootstrap 4ef8c7ec7c1df790781e"` , but get `"webpack:///webpack/bootstrap%204ef8c7ec7c1df790781e"`

redeyes2015:
About the CI failure... it is caused by an inconsistency between webpack and webpack-cli.

I filed #361 to fix it.

loganfsmyth:
@redeyes2015 FYI, I posted #367 with a more elaborate approach to using `new URL` if you're interested.

PR #361, opened 2018-09-28
Bump webpack-cli to v3.1

redeyes2015:
webpack v4.20 introduces a change making webpack-cli v2 unusable.

Check: webpack/webpack#8082.


tromey:
Seems reasonable to me.

PR #362, opened 2018-10-10
Remove the bundled dist/ directory?

loganfsmyth:
I'm in favor of removing the `dist/` folder at this point because it's not clear who the target users are, and bundling is much more common now that it was when `dist` was created in 2015. 

`source-map` already 
* requires preprocessing for older browsers since it uses `let` and `const`
* only works on browsers with WASM
* must be configured with `mappings.wasm`'s location for browsers

Those are all a complex enough that I find it hard to see the average user dropping `source-map` into their application without their own bundler. Having the extra files also increases `npm` package size with minimal benefit, and complicates maintenance if we ever want to add other dependencies since they'd all need to be bundled into `dist` too.

If we get requests for this after it is removed, I'm 100% for exploring options, but since `npm` is the primary distribution mechanism right now, it seems like most users will not be using from the registry anyway. Given that, we could also instead include standalone bundles attached to the Github releases or some other channel.


tromey:
I think this is reasonable, but TBH I don't know what the impact might be.  Also I think this should wind up in a release with an "incompatible" version bump.  Perhaps @fitzgen should weigh in here.

tromey:
Also this probably should also patch `CHANGELOG.md`.

loganfsmyth:
Yeah, I think everything for the URL changes means we'll need a new major version anyway, so that's part of why I'm trying to also post other patches that seem worth doing at the same time.

loganfsmyth:
@tromey I can't tell from your comment, do you want me to add to the CHANGELOG as part of this PR, or as part of a later update when we're ready to release?

PR #363, opened 2018-10-10
Split up wasm loading based on compilation target.

loganfsmyth:
Per my comment in https://github.com/mozilla/source-map/pull/350#issuecomment-424174926, this feels like a more ideal way to approach this. This leaves it up to bundlers to swap out the files when bundling for browser use.

I'm personally a fan of this approach because it's also an easily-reproducible check. As it is, there's no way for a module making use of `source-map` to know whether a call to `SourceMapConsumer.initialize` is needed, or will result in a `console.debug("SourceMapConsumer.initialize is a no-op when running in node.js");` message. This gives one central toggle for that, which they can also use in their own code when deciding whether to call `.initialize` or not.

### DIFF
--- a/lib/read-wasm-browser.js
+++ b/lib/read-wasm-browser.js
@@ -1,0 +1,22 @@
+"use strict";
+
+let mappingsWasm = null;
+
+module.exports = function readWasm() {
+  if (typeof mappingsWasm === "string") {
+    return fetch(mappingsWasm)
+      .then(response => response.arrayBuffer());
+  }
+  if (mappingsWasm instanceof ArrayBuffer) {
+    return Promise.resolve(mappingsWasm);
+  }
+
+  throw new Error("You must provide the string URL or ArrayBuffer contents " +
+                  "of lib/mappings.wasm by calling " +
+                  "SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) " +
+                  "before using SourceMapConsumer");
+};
+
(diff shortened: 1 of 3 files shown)


loganfsmyth:
Yup, I just posted everything I had in mind for now, but agreed that landing it after `dist` makes sense.

AdrianMrn:
I can't figure out where to find this package built for browser, could anyone point in the right direction?

loganfsmyth:
@AdrianMrn Have you potentially overridden the default `mainFields` value in Webpack? https://webpack.js.org/configuration/resolve#resolvemainfields As long as your target is `web` the `read-wasm.js` file shouldn't even be loaded: https://github.com/mozilla/source-map/blob/7adff2c69c6f2448925c94600c2eade0fbf078d7/package.json#L53 If you aren't using `browser` as a `mainFields` value that would explain the issue.

AdrianMrn:
Excuse the necro. I just picked this back up and managed to fix my issue.

It seems that, when I pulled in the package through npm, it didn't get me the latest version that has the line you mentioned, because it's a beta release. I had to manually set my version to `0.8.0-beta.0` for it to work.

yf-yang:
I encountered the same issue as AdrianMrn stated, then found I was using 0.7.4. Since 0.8.0-beta.0 has been released for 4 years without updates, is it possible to merge this commit into 0.7.x?

PR #365, opened 2018-10-15
Normalize all URLs to be as absolute as possible to avoid usage of 'util.relative'.

loganfsmyth:
The consumer currently uses `util.relative` to take a URL and convert it to one that may be present in `_sources`. There are a few downsides to this approach:

* We have to normalize `sources` or else risk not finding the match
* We have to ensure that the `sources` conform to the expectation of relative URLs
* The usage of `relative` makes it unclear whether argument URLs are expected to be relative to the map, or the root, or either.
* We have to special-case 'sources' that just aren't relative, like `sources: ['/a.js']` (which is also only special-cased on in `sourceContentFor`)

This patch centralizes all source lookups for the consumer in `_findSourceIndex` so that all APIs behave consistently. By converting all input paths to be as absolute as we can make them, we avoid the complexities of making URLs relative to one another, and it helps make clear how these behave.

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -179,11 +179,11 @@ class BasicSourceMapConsumer extends SourceMapConsumer {
       }
 
       const version = util.getArg(sourceMap, "version");
-      let sources = util.getArg(sourceMap, "sources");
+      const sources = util.getArg(sourceMap, "sources").map(String);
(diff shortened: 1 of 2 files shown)


loganfsmyth:
Woops, didn't quite post what I meant to. Closing in favor of #366

PR #371, opened 2018-11-15
Use WHATWG's URL to implement all of source-map's URL operations.

loganfsmyth:
This is #367 with a few additional commits to make things perform a bit more efficiently. The benchmark pages seem consistently on-par with the old implementation.

@jasonLaster Would you be up for reviewing the more recent 3 commits

* Only use whatwg-url in browser builds.
* Optimize perf to avoid new URL when not strictly needed.
* Cache url -> source index lookups in the consumer.


jasonLaster:
This looks good to me. Thanks for optimizing some of the hot paths. 

I hope that URL becomes more standards compliant and we can drop whatwg-url in the future

* https://bugzilla.mozilla.org/show_bug.cgi?id=1374505
* https://bugs.chromium.org/p/chromium/issues/detail?id=734880

PR #374, opened 2018-12-25
Add lastGeneratedColumn to typing for MappingItem

samccone:
This is generated when someone calls `computeColumnSpans` so we mark the property as optional as well.

### DIFF
--- a/source-map.d.ts
+++ b/source-map.d.ts
@@ -61,6 +61,7 @@ export interface MappingItem {
     source: string;
     generatedLine: number;
     generatedColumn: number;
+    lastGeneratedColumn: number | null;
     originalLine: number;
     originalColumn: number;
     name: string;


jasonLaster:
Looks good to me.

nikolay-borzov:
Any updates on this? Is there any reason why it's still pending merge?

source-map.d.ts
```diff
@@ -61,6 +61,7 @@ export interface MappingItem {
     source: string;
     generatedLine: number;
     generatedColumn: number;
+    lastGeneratedColumn?: number;
```
nikolay-borzov: I think it should be 
```
 lastGeneratedColumn: number | null;
```
If I understand code correctly the initial value is [null](https://github.com/mozilla/source-map/blob/ddaac7f00a67aefbf184bae16cb90b30ddc84a67/lib/wasm.js#L9) then it can receive number values including `Infinity`

ericmorand:
Can someone merge this and release a patch? It is a blocker for most of us using TypeScript.

jasonLaster:
Thanks @ericmorand

source-map.d.ts
jasonLaster: This makes me think it is likely optional:

https://github.com/mozilla/source-map/blob/master/lib/wasm.js#L48-L51

jasonLaster: @ericmorand @samccone which do you think is the better typing:

```js
lastGeneratedColumn?: number;
```

```js
 lastGeneratedColumn: number | null;
```

roblourens: My interpretation is that `lastGeneratedColumn: number | null;` is correct, I agree with @nikolay-borzov

`lastGeneratedColumn?: number;` is equivalent to `lastGeneratedColumn: number | undefined;` but it is initialized as `null`.

jasonLaster: Thanks. Happy to update and land.

PR #377, opened 2019-02-06
Change with type to overload

goodmind:


source-map.d.ts
```diff
@@ -236,6 +236,8 @@ export interface SourceMapConsumerConstructor {
      * console.log(xSquared);
      * ```
      */
+    with<T>(rawSourceMap: RawSourceMap, sourceMapUrl: SourceMapUrl | null | undefined, callback: (consumer: BasicSourceMapConsumer) => Promise<T> | T): Promise<T>;
+    with<T>(rawSourceMap: RawIndexMap, sourceMapUrl: SourceMapUrl | null | undefined, callback: (consumer: IndexedSourceMapConsumer) => Promise<T> | T): Promise<T>;
     with<T>(rawSourceMap: RawSourceMap | RawIndexMap | string, sourceMapUrl: SourceMapUrl | null | undefined, callback: (consumer: BasicSourceMapConsumer | IndexedSourceMapConsumer) => Promise<T> | T): Promise<T>;
```
ochameau, line 241: This is confusing. I might see an improvement if you were removing the 3rd line 241. 
So that we either type rawSourceMap as RawSourceMap or as RawIndex, but as-is the last line still allow mixed types.

PR #382, opened 2019-03-19
Treat generated column number as 0-based in indexed originalPositionFor

motiz88:
Currently `originalPositionFor` behaves differently in `BasicSourceMapConsumer` and `IndexedSourceMapConsumer`:

* In the former, the `column` argument is (correctly and as documented) treated as 0-based.
* In the latter, the `column` argument is treated as 1-based for the purpose of locating the section in which the generated position occurs (as the internal `section.generatedOffset.generatedColumn` field is, for whatever reason, 1-based).

This is a longstanding issue, going back to the introduction of `IndexedSourceMapConsumer` AFAICT, but the behavior is incorrect. This PR fixes the issue and adds a corresponding test.

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -785,7 +785,7 @@ class IndexedSourceMapConsumer extends SourceMapConsumer {
           return cmp;
         }
 
-        return (aNeedle.generatedColumn -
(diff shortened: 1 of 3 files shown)


bomsy:
this is fixed in #507

PR #398, opened 2019-06-18
`sourceContentFor(aSource, nullOnMissing=false)` should NEVER return NULL, but throw an exception instead

GerHobbelt:
when `sourceContentFor(aSource, nullOnMissing=false)` shouldn't this situation (see patch/diff) also throw an exception? In the other situation, where this API might also produce a NULL return, an exception *is* thrown instead, so this looks to me like the more consistent behaviour.

I.e. one flow path does return NULL, while the other choses NULL or exception, depending on this `nullOrMissing` parameter. Which seems odd, at least to me.

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -527,7 +527,11 @@ class BasicSourceMapConsumer extends SourceMapConsumer {
    */
   sourceContentFor(aSource, nullOnMissing) {
     if (!this.sourcesContent) {
-      return null;
+      if (nullOnMissing) {
+        return null;
+      }
+
+      throw new Error('"' + aSource + '" is not in the SourceMap.');
     }
 
     const index = this._findSourceIndex(aSource);
--- a/lib/source-map-generator.js
+++ b/lib/source-map-generator.js
@@ -79,7 +79,7 @@ class SourceMapGenerator {
         generator._sources.add(sourceRelative);
       }
 
-      const content = aSourceMapConsumer.sourceContentFor(sourceFile);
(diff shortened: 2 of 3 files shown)


GerHobbelt:
simile of #391 😢

andrewnicols:
Heh. Patches are almost identical except for unit tests. Great minds think alike and all that ;)

GerHobbelt:
> Heh. Patches are almost identical except for unit tests. Great minds think alike and all that ;)

Nah. It's rather more embarrassing: I had merged your code in my own fork, went on debugging some other other stuff in this lib and then when that was completed successfully, I did a manual inspection diff with the mainline, producing this one as one of the important changes. While I had been agonizing over the bugs elsewhere, I had completely forgotten about this merge.

Only after filing the pullreq and going through the other pending pullreqs did I notice I had picked up your code. Hence the edited-in remark above:

> 😢 Should better check/document my merges.

The only remaining worth is the extra unit tests.

andrewnicols:
Sorry to break it to you, but this diffset has fewer tests than my issue. You’ve removed the ones in my patchset. At least if I read the changes tab properly.

GerHobbelt:
Did a double-check: yes, you are absolutely correct. 😓   Thought I had it the other way around.

Killing this one, as it's a copy of #391 and y'all should ride with that one.

GerHobbelt:
BTW, @andrewnicols : thanks for pointing this out; I had it completely wrong in my mind.

GerHobbelt:
Dupe of #391

PR #399, opened 2019-06-18
Off by one(1) error in column addressing: fixing `originalPositionFor()` API for indexed maps + added test adapted from the 0.6.x dev tree

GerHobbelt:
The main work was done in SHA-1: 88efe6141cbf73c11170eba9e8696c7931554f6f

* fixing `originalPositionFor()` API for indexed maps + added test adapted from the 0.6.x dev tree: it turned out that due an off-by-1 error in the column value, half of the mapping nodes weren't hit by the `originalPositionFor()` internal search code, while it *should* find those.

The added tests are augmented with extra logging for diagnosis: run those tests with & without the given patch to observe the bug & change.

---

Before this pullreq, the 0.7.x/0.8.x did not test this situation; I ran into trouble with indexed maps crashing and acting crazy. The added test ensures there's no regression.

**WARNING / NOTE**: this pullreq includes #394 and #398 as required prerequisites: without those, this won't fly, as the crashes/failures in the test set will simply move to other spots in the code.
Hence this is more complex issue as #394 and #398 must be fixed before this one can work.

See the commit log for this one: patch-3 = #394 and patch-7 = #398 which are visible below as `git merge` commits.

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -785,8 +785,7 @@ class IndexedSourceMapConsumer extends SourceMapConsumer {
           return cmp;
         }
 
-        return (aNeedle.generatedColumn -
-                section.generatedOffset.generatedColumn);
+        return aNeedle.generatedColumn - (section.generatedOffset.generatedColumn - 1);
       });
     const section = this._sections[sectionIndex];
 
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -527,7 +527,11 @@ class BasicSourceMapConsumer extends SourceMapConsumer {
    */
   sourceContentFor(aSource, nullOnMissing) {
     if (!this.sourcesContent) {
-      return null;
+      if (nullOnMissing) {
+        return null;
+      }
+
+      throw new Error('"' + aSource + '" is not in the SourceMap.');
     }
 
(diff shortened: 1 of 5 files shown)


GerHobbelt:
Ha! This is a very close simile of #382: I browsed the pullreq list just now and discovered I had ripped that one's unit test into my 0.6.x and then went for 0.8.x and redid the work.
Turns out my edit is -1 decrement of the rightmost term, while #382 @motix88 had +1 increment for the leftmost term, making these fixes work the same.

However 😉 , the -1 approach is already present a little lower in the code in this little chunk that sits just below the place where we both patched the code:

```
    return section.consumer.originalPositionFor({
      line: needle.generatedLine - (section.generatedOffset.generatedLine - 1),
// ^^^^^^^^^^^^^^^^^^^^ note this existing line ^^^^^^^^^^
      column:
        needle.generatedColumn -
        (section.generatedOffset.generatedLine === needle.generatedLine
          ? section.generatedOffset.generatedColumn - 1
          : 0),
      bias: aArgs.bias
    });
```

Also, this pullreq makes all tests pass and for that I had to mix in the other pullreqs mentioned earlier. Without them, the test kept failing; see also the console diagnostics in the commits.

bomsy:
Closing as fixed by #507

PR #404, opened 2019-08-18
readme: use `const {SourceNode, SourceMapConsumer, SourceMapGenerator…

zaoqi:
readme: use `const {SourceNode, SourceMapConsumer, SourceMapGenerator} = `

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,13 +184,13 @@ Get a reference to the module:
 
 ```js
 // Node.js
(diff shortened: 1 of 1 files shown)


jkrems:
Thanks for the suggestion! While I agree that this style is more typically used, the code samples that follow all use `sourceMap.*` when referencing the interface which would be confusing with this change.

Sorry for the slow response here.

Issue #400, opened 2019-06-26
Consider removing whatwg-url polyfill


Rich-Harris:
Hi all. We're considering adopting source-map in a project that will be used in browsers, and this chart is giving us pause:

<img width="946" alt="Screen Shot 2019-06-26 at 10 40 52" src="https://user-images.githubusercontent.com/1162160/60189442-e91d2700-97fe-11e9-9f58-d0f550cbcd5e.png">

* **0.7.3**: 28.2kb minified, 8.1kb zipped
* **0.8.0-beta.0**: 377.3kb minified, 95kb zipped

This stark difference is entirely accounted for by the addition of [whatwg-url](https://www.npmjs.com/package/whatwg-url) as a dependency. It doesn't affect Node at all (which either uses the built in `URL` object or `require("url").URL` as appropriate), so this polyfill is for the benefit of browsers that don't support `URL`, which is [basically just IE and Opera Mini](https://caniuse.com/#feat=url).

If this package is intended to be used in browsers, I'd argue that keeping the bundle size small is a worthwhile goal. I see a couple of options:

* Expect developers serving legacy browsers to bring their own polyfill
* Reimplement the same functionality used by this package in a small utility module

I'd be happy to work on a pull request if you could give me some guidance. Thank you!

Teoxoy:
They mention here why `whatwg-url` was added

https://github.com/mozilla/source-map/blob/02f3d6bbcd4cb2c15dc7083d9afb561879ad2b25/lib/url-browser.js#L9-L20

PR #413, opened 2019-12-17
chore(browser): swap URL polyfill

lukeed:
Hey there,

I put together a [1.5kB polyfill](https://github.com/lukeed/url-shim) specifically to unblock this issue: https://github.com/mozilla/source-map/issues/400

We'd really like to move forward with `source-map` integration within [the Svelte REPL environment](https://svelte.dev/repl/hello-world?version=3.16.4), but not at the expense of nearly doubling its javascript! With `whatwg-url` included – which is [rarely needed](https://caniuse.com/#feat=url) (especially on a developer machine) – `source-map` weighs 30% _more_ than CodeMirror.

This polyfill, like `whatwg-url`, polyfills the `URL` implementation from *Node.js* instead of the browser's. It also passes the linked Chromium & Mozilla issues.

### DIFF
--- a/lib/url-browser.js
+++ b/lib/url-browser.js
@@ -18,4 +18,4 @@
  * * https://bugzilla.mozilla.org/show_bug.cgi?id=1374505
  * * https://bugs.chromium.org/p/chromium/issues/detail?id=734880
  */
-module.exports = require("whatwg-url").URL;
+module.exports = require("url-shim").URL;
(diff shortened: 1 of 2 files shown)


lukeed:
Hey :) I know we're all busy with the holidays, but just wanted to check in. If this has no chance of landing, that's totally fine, but please let me know so that our side can start formulating backup plans in order to move forward.

Thanks! ☃️

nchevobbe:
We'll drop the `whatwg-url` module in #400

PR #445, opened 2021-08-14
Fix `package.json#typings` field

chalkygames123:
Fixes: https://stackoverflow.com/questions/63124462/how-to-fix-file-node-modules-dotenv-types-not-found-error-coming-from-j

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
   "nyc": {
     "reporter": "html"
   },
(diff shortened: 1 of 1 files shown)


ochameau:
`typings` is redundant with `types` attribute, which should be correct.
#448 removed `typings`.
Let me know if you still face issues around typescript.

PR #448, opened 2021-11-22
Include types file extension in package.json

amcasey:
Otherwise, the TypeScript language service used in VS Code (and other editors) won't pick up the types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
   "nyc": {
     "reporter": "html"
   },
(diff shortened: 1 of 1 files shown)


ochameau (approved):
I wasn't able to reproduce issues with VS.code language service, but yes `typings` is redundant with `types`:
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

PR #463, opened 2022-10-06
Update ESLint & Prettier to latest

eemeli:
This should make it easier to work with the project, as atm it appears to require using `npm i --legacy-peer-deps` to get running.

ESLint rules that are covered by Prettier are dropped, and a `"prettier"` config is added to the package.json to account for the default changes in its v1 -> v2 transition.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,6 @@ module.exports = {
     es6: true
   },
 
-  plugins: ["prettier"],
   extends: ["prettier"],
 
   parserOptions: {
@@ -19,30 +18,12 @@ module.exports = {
   },
 
   rules: {
-    // Require spacing around =>
-    "arrow-spacing": "error",
-
-    // Braces only needed for multi-line arrow function blocks
-    // "arrow-body-style": ["error", "as-needed"]
-
     // Always require spacing around a single line block
     "block-spacing": "error",
-
-    // No newline before open brace for a block
-    "brace-style": ["error", "1tbs", { allowSingleLine: true }],
-
-    // No space before always a space after a comma
-    "comma-spacing": ["error", { after: true, before: false }],
-
-    // Commas at the end of the line not the start
-    "comma-style": "error",
 
     // Warn about cyclomatic complexity in functions.
     // XXX Get this down to 20?
     complexity: ["error", 25],
-
-    // Don't require spaces around computed properties
-    "computed-property-spacing": ["error", "never"],
 
     // Functions must always return something or nothing
     "consistent-return": "error",
@@ -55,28 +36,6 @@ module.exports = {
 
     // Always require a trailing EOL
     "eol-last": "error",
-
-    // No spaces between function name and parentheses
-    "func-call-spacing": "error",
-
-    // Require function* name()
-    "generator-star-spacing": ["error", { after: true, before: false }],
-
-    // Two space indent
-    // "indent": ["error", 2, { "SwitchCase": 1 }],
-
-    // Space after colon not before in property declarations
-    "key-spacing": [
-      "error",
-      {
-        afterColon: true,
-        beforeColon: false,
-        mode: "minimum"
-      }
-    ],
-
-    // Require spaces before and after keywords
-    "keyword-spacing": "error",
 
     // Unix linebreaks
     "linebreak-style": ["error", "unix"],
@@ -151,9 +110,6 @@ module.exports = {
     // No using !! where casting to boolean is already happening
     "no-extra-boolean-cast": "error",
 
-    // No double semicolon
-    "no-extra-semi": "error",
-
     // No overwriting defined functions
     "no-func-assign": "error",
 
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -19,31 +19,16 @@ module.exports = {
...
-    "arrow-spacing": "error",
-
-    // Braces only needed for multi-line arrow function blocks
-    // "arrow-body-style": ["error", "as-needed"]
-
     // Always require spacing around a single line block
     "block-spacing": "error",
 
     // No newline before open brace for a block
     "brace-style": ["error", "1tbs", { allowSingleLine: true }],
```
jkrems, line 26: Should this be covered by prettier as well? I assume it would move the brace around?

.eslintrc.js
```diff
@@ -239,9 +196,6 @@ module.exports = {
     // Disallow tabs.
     "no-tabs": "error",
```
jkrems, line 197: Would tabs already be replaced by prettier?

jkrems (approved):
Skimmed the formatting changes, RSLGTM.

.eslintrc.js
eemeli: On further investigation, this should be dropped as well. This rule does match what Prettier does, but it's useless.

.eslintrc.js
eemeli: There are places like comments where Prettier doesn't touch tabs, while this rule would complain about them.

.eslintrc.js
eemeli, line 26:
```suggestion
```

eemeli:
Rebased on `master` to trigger CI jobs.

package.json
```diff
@@ -74,24 +74,28 @@
     "dev:watch": "watch 'npm run coverage' lib/ test/",
     "predev": "npm run setup",
     "dev": "npm-run-all -p --silent dev:*",
-    "prettier": "prettier --write --print-width=120 *.{js,ts,json,md} lib/*.js test/*.js bench/bench*",
```
eemeli, line 77: During the rebase I noticed that there's a print-width setting here that isn't actually followed for all files; the diff was actually smaller when going with the default 80, so I did that.

ochameau (approved):
Thanks a lot for looking into this!

Note that when running `npm run prettier` locally I see modifications against [.waiting.html](https://github.com/mozilla/source-map/blob/master/.waiting.html). It might relate to your very latest change.
Shouldn't we have a check against prettier on Github actions?
It looks like [we only run eslint](https://github.com/mozilla/source-map/blob/master/.github/workflows/nodejs.yml#L33-L34).
Would it be easy to run prettier and fail in case of modification?

> This should make it easier to work with the project, as atm it appears to require using `npm i --legacy-peer-deps` to get running.

This is for my own culture (feel free to ignore), but for me it works without `--legacy-peer-deps`.
```
$ npm i
npm WARN eslint-plugin-prettier@3.4.1 requires a peer of eslint@>=5.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.13 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.13: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

removed 157 packages and audited 779 packages in 4.887s

62 packages are looking for funding
  run `npm fund` for details

found 98 vulnerabilities (17 moderate, 59 high, 22 critical)
  run `npm audit fix` to fix them, or `npm audit` for details

$ npm run lint
=> WORKS
$ npm run prettier
=> ALSO WORKS
```
As if the warning during `npm i` is harmless?
(I'm on npm 6.14.17 and node v14.20.1)

Otherwise I confirm that this prettier error during `npm i` disappear with your patch.

> ESLint rules that are covered by Prettier are dropped, and a `"prettier"` config is added to the package.json to account for the default changes in its v1 -> v2 transition.

I tried to find what rules prettier was applying without much success. Were you able to find such list online?
Again, this isn't super important as your patch looks good as-is.

I'd be more interested in trying to align slightly more with mozilla-central linting.
Regarding prettier rules, I see the following failures:
```
lib/wasm.js
  115:12  error  Insert `,`                                     prettier/prettier (eslint)
  116:10  error  Insert `,`                                     prettier/prettier (eslint)
  129:10  error  Insert `,`                                     prettier/prettier (eslint)
source-map.js
   6:29  error  Delete `⏎·`                                    prettier/prettier (eslint)
   8:28  error  Delete `⏎·`                                    prettier/prettier (eslint)
and many others...
```

Otherwise beyond prettier, there is a few rules on mozilla-central that fails:
```
Use the global form of 'use strict'.                                                       strict (eslint)
Parameter 'aArgs' uses Hungarian Notation, consider using 'args' instead.                  mozilla/no-aArgs (eslint)
use .includes instead of .indexOf                                                          mozilla/use-includes-instead-of-indexOf (eslint)
Expected { after 'if' condition.                                                           curly (eslint)
error  Prefer boolean length check                                                                mozilla/prefer-boolean-length-check (eslint)
```
We can also look into that in followups and not necessarily align all rules.

PR #464, opened 2022-10-06
Require Node.js 12 or later & switch CI to GitHub Actions

eemeli:
Really we could update the minimum dependency to Node.js 14 as everything older than that is out of support.

The CI tests are here set up to run on a 4x3 matrix of systems, covering Node.js 12, 14, 16, and 18 on Linux, Mac & Windows.

As the action doesn't get run for the PR actually adding it (that's controlled by the config on the default branch), a test run is available here: https://github.com/eemeli/source-map/actions/runs/3196070788


hildjj:
+1 for testing & supporting node [14,16,18] only.  That's a semver_major change, in my opinion.

hildjj (approved):

hildjj:
Version numbers are cheap. :)
My opinion is start on 1.0 now, then use 2.0 (or whatever is next at that point) for the refactor.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,39 @@
...
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        node-version: [12, 14, 16, 18]
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm install --legacy-peer-deps
```
jkrems, line 24: Is it worth it to create a `package-lock.json` and use `npm ci` here to ensure reproducible builds?

eemeli: Eventually, yes. I'm hoping to add a separate PR adding a `package-lock.json` file, once at least this and #463 have been merged.

jkrems (approved):

ochameau (approved):
Thanks for jumping on this right away, this looks good.
See my inline comments.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,39 @@
...
+      - run: npm test
+
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+      - run: npm install --legacy-peer-deps
```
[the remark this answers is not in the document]
eemeli: Yes, and yes.

eemeli:
Tested the merge on my fork, looks pretty green: https://github.com/eemeli/source-map/actions/runs/3295739934

PR #467, opened 2022-11-03
Adopt & enforce the mozilla-central Prettier config

eemeli:
As discussed in #463, this applies the current mozilla-central [Prettier config](https://searchfox.org/mozilla-central/rev/49011d374b626d5f0e7dc751a8a57365878e65f1/.prettierrc) to this repository, and adds a check for it in the CI.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,18 +3,18 @@
 module.exports = {
   env: {
     node: true,
-    es6: true
+    es6: true,
   },
 
   extends: ["prettier"],
 
   parserOptions: {
-    ecmaVersion: 9
+    ecmaVersion: 9,
   },
 
   globals: {
     fetch: false,
-    WebAssembly: false
+    WebAssembly: false,
   },
 
   rules: {
@@ -142,9 +142,9 @@ module.exports = {
           ArrayExpression: true,
           AssignmentExpression: true,
           ObjectExpression: true,
-          VariableDeclarator: true
-        }
-      }
+          VariableDeclarator: true,
+        },
+      },
     ],
 
     // No reassigning native JS objects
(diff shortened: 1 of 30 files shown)


lib/read-wasm-browser.js
```diff
@@ -18,6 +18,6 @@ module.exports = function readWasm() {
   );
 };
 
-module.exports.initialize = input => {
+module.exports.initialize = (input) => {
```
ochameau, line 21: I'm not able to understand why... but on mozilla-central we don't add parens on single arguments.

I saw https://github.com/prettier/prettier/pull/3324 but I don't see allowParens usages in mozilla-central...

ochameau (approved):
Thanks for jumping on this followup!

It looks good to me, there is just this one difference with m-c regarding single argument parenthesis for arrow functions.

lib/read-wasm-browser.js
eemeli: m-c is using [`prettier@1.19.1`](https://searchfox.org/mozilla-central/rev/0b1543e85d13c30a13c57e959ce9815a3f0fa1d3/package.json#28), and the default value for `allowParens` [changed](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev) in the 2.0 release.

@ochameau Should we add `arrowParens: avoid` to our config to match the previous behaviour?

ochameau: Oh nice catch! (I'm so happy to have you around so knowledgable about all this!!)

Yes, it would be nice. This would avoid having some useless changes when updating source-map code in -mc.

eemeli:
Rebased to add `arrowParens: avoid` to the config & add the PR id directly to the commit headers so that we can "Rebase and merge" rather than squash the commits to keep the config change & its effects separate.

PR #335, opened 2018-05-05
Coverage tools

hildjj:
This is just a suggestion, but I've used a similar approach on several projects and really like how it drives coverage up.

Rebased on #332.  Adds support for a developer workflow kicked off with `npm run dev`, which will:

- Kick off a webserver
- Run coverage metrics
- Open the metrics in the default browser
- On file change, re-run metrics, and update metrics in browser

Also add support for reporting coverage on travis builds.  For an example, see
https://coveralls.io/github/hildjj/source-map

To enable coveralls for the mozilla branch, go to https://coveralls.io/repos/new , log in with GitHub, find source-map, and click the enable button.  I think I may have the necessary privileges to do that for us, needed.

PR #468, opened 2022-11-03
Simplify coverage tooling

eemeli:
This simplifies the coverage reporting tooling that was added in #335. The dev dependencies added there have accrued a number of vulnerabilities, and are a bit excessive compared to the complexity they introduce.

With these changes, the `npm run coverage` script will still produce an HTML coverage report at `coverage/index.html`, but it'll also print the results on the command line. The web server for this is removed, as are the scripting tools required for live-updating its view.


ochameau (approved):
This all looks good to me.
I agree that the complexity added for live-update of codecoverage is probably not worth the cost.

Do you think we could easily revive codecoverage report on pull requests?
I'm not sure it is reasonable to ask to enable coveralls github action on all mozilla repos, but may be we can do something manual?

eemeli:
> @ochameau:
> Do you think we could easily revive codecoverage report on pull requests?
> I'm not sure it is reasonable to ask to enable coveralls github action on all mozilla repos, but may be we can do something manual?

It looks like the `codecov/codecov-action` is already approved, so using that could be an option. There are also a number of other solutions for GH actions. I think a key step here would be to determine what's the expected/desired benefit of code coverage reporting.

eemeli:
Rebased on master to resolve nominal conflicts.

PR #469, opened 2022-11-09
Update to doctoc v2 & simplify call pattern

eemeli:
This also adds a check to the lint task to validate the current TOCs.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -31,6 +31,7 @@ jobs:
       - uses: actions/setup-node@v3
       - run: npm install --legacy-peer-deps
       - run: npm run lint
+      - run: npm run toc
       - run: npx prettier --check .
       - run: npm run coverage
       - run: npx nyc report --reporter=text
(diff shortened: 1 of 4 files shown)


.github/workflows/nodejs.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       - uses: actions/setup-node@v3
       - run: npm install --legacy-peer-deps
       - run: npm run lint
+      - run: npm run toc
```
ochameau, line 34: Out of curiosity, would it fail if any modification happens?

eemeli:
> @ochameau:
> Note that I see local differences when running `npm run toc`: [...]

Did you run `npm install` to update doctoc, or is that with its latest version?

.github/workflows/nodejs.yml
eemeli: Good point; no, it won't. I'd somehow gotten it into my head that a dirty worktree would cause a failure somewhere in this task, but clearly that's not the case. Will fix.

PR #470, opened 2022-11-09
Filter out unnecessary rust tasks

eemeli:
1. We shouldn't prettify stuff in `wasm-mappings/target/`.
2. We don't need to run the heavier rust tests if we don't change any of the relevant code.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -1,6 +1,9 @@
 on:
   pull_request:
     branches: [master]
+    paths:
+      - wasm-mappings/**
+      - .github/workflows/cargo.yml
   push:
     branches: [master]
   workflow_dispatch:
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@
 bench/
 coverage/
 node_modules/
+target/


.github/workflows/cargo.yml
```diff
@@ -1,6 +1,9 @@
 on:
   pull_request:
     branches: [master]
+    paths:
+      - wasm-mappings/**
+      - .github/workflows/cargo.yml
```
ochameau, line 6: Oh, thanks I was wondering how to do that!

.prettierignore
```diff
@@ -2,3 +2,4 @@
 bench/
 coverage/
 node_modules/
+target/
```
ochameau, line 5: Did you mean `wasm-mappings/target/`?

You may ignore the whole `wasm-mappings/` as there is also json files being pretty printed in `wasm-mappings/source-map-mappings-wasm-api/`.

eemeli: Without a leading `/`, this'll match any directory named `target`, so it'll take care of `wasm-mappings/source-map-mappings-wasm-api/target/` as well.

PR #471, opened 2022-11-09
Improve coverage

eemeli:
As a follow-on to #468, this switches the coverage tool from `nyc` to `c8`, which use's Node's built-in coverage tooling to generate compatible results.

Those results are then made visible via codecov.io, where I've enabled reporting for the `mozilla/source-map` repo. I'm not sure about their GH bot having access to writing statuses & comments to PRs yet, though.

As a side effect, this gets rid of all the remaining known vulnerable npm dependencies.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -35,3 +35,5 @@ jobs:
       - run: git diff --stat --exit-code
       - run: npx prettier --check .
       - run: npm run coverage
+      - run: npx c8 report --reporter=lcov
+      - uses: codecov/codecov-action@v3
--- a/package.json
+++ b/package.json
@@ -67,16 +67,16 @@
   "scripts": {
     "lint": "eslint --fix *.js lib/ test/",
     "test": "node test/run-tests.js",
-    "coverage": "nyc --reporter=text --reporter=html npm test",
+    "coverage": "c8 --reporter=text --reporter=html npm test",
     "prettier": "prettier --write .",
-    "clean": "rm -rf coverage .nyc_output",
+    "clean": "rm -rf coverage",
     "toc": "doctoc --github --notitle README.md CONTRIBUTING.md"
   },
   "devDependencies": {
+    "c8": "^7.12.0",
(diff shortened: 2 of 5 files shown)


eemeli:
Well, looks like a comment is at least visible here. It's [configurable](https://docs.codecov.com/docs/pull-request-comments#layout).

Are we happy with that, or do we want something a bit more discreet via [commits status checks](https://docs.codecov.com/docs/commit-status)?

.github/workflows/nodejs.yml
```diff
@@ -35,3 +35,5 @@ jobs:
       - run: git diff --stat --exit-code
       - run: npx prettier --check .
       - run: npm run coverage
+      - run: npx c8 report --reporter=lcov
+      - uses: codecov/codecov-action@v3
```
ochameau, line 39: Don't you need to pass the location of coverage files?
Is it magic enough to find them in coverage folder?

package.json
```diff
@@ -67,16 +67,16 @@
...
     "toc": "doctoc --github --notitle README.md CONTRIBUTING.md"
   },
   "devDependencies": {
+    "c8": "^7.12.0",
     "doctoc": "^2.2.1",
     "eslint": "^8.24.0",
     "eslint-config-prettier": "^8.5.0",
-    "nyc": "^11.7.1",
     "prettier": "^2.7.1"
   },
```
ochameau, line 81: Nice job on getting the dev dependencies all updated and renewed to best pratices!

ochameau (approved):
> As a side effect, this gets rid of all the remaining known vulnerable npm dependencies.

Nice work to get to this!

> Well, looks like a comment is at least visible here. It's [configurable](https://docs.codecov.com/docs/pull-request-comments#layout).
> 
> Are we happy with that, or do we want something a bit more discreet via [commits status checks](https://docs.codecov.com/docs/commit-status)?

I like the PR comment, you get an immediate input on improvements/regressions.
I think commits status would only work if we set some strict minimum threshold which can be painful on the long run.
I may only suggest enabling the `require_changes: true` option.

PR #489, opened 2023-05-28
ci: use setup-node caching

nschonni:
The setup-node action supports built-in package caching since v2


eemeli (approved):
This seems pretty safe and sensible.

PR #490, opened 2023-05-28
chore: don't fix with lint target

nschonni:
Since the `lint` target is called with the `--fix` parameter, the CI wouldn't fail if there were autofixable issues. Split the fix vs lint targets, and reused Prettiers ignore file instead of creating a `.eslintignore` file

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,8 @@
   },
   "license": "BSD-3-Clause",
   "scripts": {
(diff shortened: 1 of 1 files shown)


eemeli (changes requested):
Rather than changing the behaviour of `npm run lint`, it'd be better to change the command used in CI to directly call `npx eslint` without the `--fix` argument.

PR #491, opened 2023-05-29
chore: various typos

nschonni:
Ran accross the repo with cSpell


bench/self-source-map.js
eemeli, line 1: It would be better to leave this file untouched, as it's only used for benchmarking. Also, I can't even spot what the changes here are.

eemeli (changes requested):
Thank you for spotting these! One small fix, and I think this is good to merge.

bench/self-source-map.js
nschonni: It was a find/replace for minimium/minimum like in a few of the other files

PR #494, opened 2023-07-06
Fix a few typos with codespell

alexrudd2:
Happened to be running `codespell` against one of my repos too broadly, 
and it caught some typos here.

(I didn't change `self-source-map.js`, since I assume that's automatically 
generated)


alexrudd2:
Oh, might be redundant with https://github.com/mozilla/source-map/pull/494.  Sorry!

jkrems:
#491 is likely what you meant to link. :)

Issue #504, opened 2024-04-19
Start running source map specification tests


bomsy:
Work on the source map spec tests is in progress. It currently in https://github.com/takikawa/source-map-tests. Lets add the infrastructure to start running the tests from here.

PR #505, opened 2024-04-27
Add source map spec tests

takikawa:
This is a draft PR for issue #504.

The PR adds integration to run the in-development source map specification tests. The tests are accessed via a git submodule, though the current location of the tests is likely not their final official location. The tests are still under development as well.

### DIFF
--- a/test/test-spec-tests.js
+++ b/test/test-spec-tests.js
@@ -1,0 +1,102 @@
+/* -*- Mode: js; js-indent-level: 2; -*- */
+/*
+ * Copyright 2024 Mozilla Foundation and contributors
+ * Licensed under the New BSD license. See LICENSE or:
+ * http://opensource.org/licenses/BSD-3-Clause
+ */
+
+const fs = require('node:fs/promises');
+const SourceMapConsumer =
+  require("../lib/source-map-consumer").SourceMapConsumer;
+
+const sourceMapSpecTests = require("./source-map-tests/source-map-spec-tests.json");
+
+async function readJSON(path) {
+  const file = await fs.open(require.resolve(path));
+  const json = JSON.parse(await file.readFile());
+  file.close();
+  return json;
+}
+
+// Known failures due to intentional implementation choices or due to bugs.
+const skippedTests = [
+  // Versions are explicitly checked a bit loosely.
+  "versionNumericString",
+  // Stricter sources array checking isn't implemented.
+  "sourcesNotStringOrNull",
+  "sourcesAndSourcesContentBothNull",
+  // Stricter names array checking isn't implemented.
+  "namesMissing",
+  "namesNotString",
+  // This check isn't as strict in this library.
+  "invalidMappingNotAString1",
+  // A mapping segment with no fields is technically invalid in the spec.
+  "invalidMappingSegmentWithZeroFields",
+  // These tests fail due to imprecision in the spec about the 32-bit limit.
+  "invalidMappingSegmentWithColumnExceeding32Bits",
+  "invalidMappingSegmentWithOriginalLineExceeding32Bits",
+  "invalidMappingSegmentWithOriginalColumnExceeding32Bits",
+  // A large VLQ that should parse, but currently does not.
+  "validMappingLargeVLQ",
+  // The library currently doesn't check the types of offset lines/columns.
+  "indexMapOffsetLineWrongType",
+  "indexMapOffsetColumnWrongType",
+  // The spec is not totally clear about this case.
+  "indexMapInvalidBaseMappings",
+  // The spec's definition of overlap can be refined
+  "indexMapInvalidOverlap",
+  // Not clear if this test makes sense, but spec isn't clear on behavior
+  "validMappingNullSources"
+]
+
+async function testMappingAction(assert, rawSourceMap, action) {
+  return SourceMapConsumer.with(rawSourceMap, null, (consumer) => {
+    let mappedPosition = consumer.generatedPositionFor({
+      source: action.originalSource,
+      line: action.originalLine + 1,
+      column: action.originalColumn
+    });
+
+    assert.equal(mappedPosition.line, action.generatedLine + 1, `generated line didn't match, expected ${action.generatedLine + 1} got ${mappedPosition.line}`);
+    assert.equal(mappedPosition.column, action.generatedColumn, `generated column didn't match, expected ${action.generatedColumn} got ${mappedPosition.column}`);
+
+    mappedPosition = consumer.originalPositionFor({
+      line: action.generatedLine + 1,
+      column: action.generatedColumn,
+    });
+
+    assert.equal(mappedPosition.line, action.originalLine + 1, `original line didn't match, expected ${action.originalLine + 1} got ${mappedPosition.line}`);
+    assert.equal(mappedPosition.column, action.originalColumn, `original column didn't match, expected ${action.originalColumn} got ${mappedPosition.column}`);
+    assert.equal(mappedPosition.source, action.originalSource, `original source didn't match, expected ${action.originalSource} got ${mappedPosition.source}`);
+    if (action.mappedName)
+      assert.equal(mappedPosition.name, action.mappedName, `mapped name didn't match, expected ${action.mappedName} got ${mappedPosition.name}`);
+  });
+}
+
+for (let testCase of sourceMapSpecTests.tests) {
+  if (skippedTests.includes(testCase.name))
+    continue;
+  exports[`test from source map spec tests, name: ${testCase.name}`] =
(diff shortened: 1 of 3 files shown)


test/test-spec-tests.js
```diff
@@ -0,0 +1,91 @@
...
+  // The library currently doesn't check the types of offset lines/columns.
+  "indexMapOffsetLineWrongType",
+  "indexMapOffsetColumnWrongType",
+  // The spec is not totally clear about this case.
+  "indexMapInvalidBaseMappings",
+  // The spec's definition of overlap can be refined
+  "indexMapInvalidOverlap",
+  // Not clear if this test makes sense, but spec isn't clear on behavior
+  "validMappingNullSources"
+]
```
bomsy, line 50: Are these skipped tests specific to the mozilla sourcemap implementation
```suggestion
// Known failures due to intentional implementation choices or due to bugs.
const skippedTests = [
  // Versions are explicitly checked a bit loosely.
  "versionNumericString",
  // Stricter sources array checking isn't implemented.
  "sourcesNotStringOrNull",
  "sourcesAndSourcesContentBothNull",
  // Stricter names array checking isn't implemented.
  "namesMissing",
  "namesNotString",
  // This check isn't as strict in this library.
  "invalidMappingNotAString1",
  // A mapping segment with no fields is technically invalid in the spec.
  "invalidMappingSegmentWithZeroFields",
  // These tests fail due to imprecision in the spec about the 32-bit limit.
  "invalidMappingSegmentWithColumnExceeding32Bits",
  "invalidMappingSegmentWithOriginalLineExceeding32Bits",
  "invalidMappingSegmentWithOriginalColumnExceeding32Bits",
  // A large VLQ that should parse, but currently does not.
  "validMappingLargeVLQ",
  // The library currently doesn't check the types of offset lines/columns.
  "indexMapOffsetLineWrongType",
  "indexMapOffsetColumnWrongType",
  // The spec is not totally clear about this case.
  "indexMapInvalidBaseMappings",
  // The spec's definition of overlap can be refined
  "indexMapInvalidOverlap",
  // Not clear if this test makes sense, but spec isn't clear on behavior
  "validMappingNullSources"
]
```

test/test-spec-tests.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+}
+
+for (let testCase of sourceMapSpecTests.tests) {
+  if (skippedTests.includes(testCase.name))
+    continue;
+  exports[`test from source map spec tests, name: ${testCase.name}`] =
+    async function (assert) {
+      const json = await readJSON(`./source-map-tests/resources/${testCase.sourceMapFile}`);
+      let sourceMapFailed = false;
```
bomsy, line 71: This does not seem to be used!

bomsy:
Thanks for this patch. this is looking great. I just had a quick look. I'll be testing it in a little bit

takikawa:
Note: I updated the patch with support for checking generated->original mapping (before it just checked original->generated). Now it depends on PR #507 as well.

I also updated it with checking for transitive mapping. Both of these are pretty localized changes.

test/test-spec-tests.js
takikawa: Thanks for catching that, I think it was a vestige of an earlier attempt. I'll remove it next time I update the patch.

test/test-spec-tests.js
takikawa: Yes, for example the "versionNumericString" test is skipped because this library specifically accepts the case where the version field is a number in a string. Other implementations have different behavior (e.g., the [source map validator](https://github.com/jkup/source-map-validator) uses this library but layers more strict checking on top).

test/test-spec-tests.js
```diff
@@ -0,0 +1,129 @@
...
+      try {
+        const map = await new SourceMapConsumer(json);
+        map.eachMapping(() => {});
+        map.destroy();
+      } catch (exn) {
+        if (testCase.sourceMapIsValid)
```
bomsy, line 113:
```suggestion
        if (testCase.sourceMapShouldBeValid)
```

bomsy: This is not related directly to this patch but it just a tiny suggestion while trying the understanding the code.
`sourceMapShouldBeValid` might make it clearer what the expectation is.

test/test-spec-tests.js
```diff
@@ -0,0 +1,129 @@
...
+          assert.fail("Expected valid source map but failed to load successfully: " + exn.message);
+        return;
+      }
+      if (!testCase.sourceMapIsValid)
+        assert.fail("Expected invalid source map but loaded successfully");
+      if (testCase.testActions) {
+        for (let testAction of testCase.testActions) {
```
bomsy, line 120:
```suggestion
        for (const testAction of testCase.testActions) {
```

bomsy: Maybe we could use `const` since `testAction` is not reassigned.

test/test-spec-tests.js
```diff
@@ -0,0 +1,129 @@
...
+async function testTransitiveMappingAction(assert, rawSourceMap, action) {
+  return SourceMapConsumer.with(rawSourceMap, null, async (consumer) => {
+    assert.ok(Array.isArray(action.intermediateMaps), "transitive mapping case requires intermediate maps");
+
+    let mappedPosition = consumer.originalPositionFor({
+      line: action.generatedLine + 1,
+      column: action.generatedColumn,
+    });
+
+    for (let intermediateMapPath of action.intermediateMaps) {
```
bomsy, line 85:
```suggestion
    for (const intermediateMapPath of action.intermediateMaps) {
```

bomsy: Since `intermediateMapPath` is not reassigned

test/test-spec-tests.js
```diff
@@ -0,0 +1,129 @@
...
+  "indexMapInvalidOverlap",
+  // Not clear if this test makes sense, but spec isn't clear on behavior
+  "validMappingNullSources"
+]
+
+async function testMappingAction(assert, rawSourceMap, action) {
+  return SourceMapConsumer.with(rawSourceMap, null, (consumer) => {
+    let mappedPosition = consumer.generatedPositionFor({
+      source: action.originalSource,
+      line: action.originalLine + 1,
```
bomsy, line 56:
```suggestion
      line: mapLine(action.originalLine),
```
```
function mapLine(line) {
  return line + 1;
}
```

bomsy: So we seem to be adding `1` to all the lines in most places here, we can have a util function which adds the line.

bomsy:
Thanks @takikawa. I tested out the patch and it works nicely. I've added a few inline comments.

Once the tests have been moved to the main repo we can updated and land this patch. Do you know the timeline for this?

Also could we automate the update of the git sub-module by  adding the command to the npm task for running tests in the `package.json`
This simple task seems to work for me. 
```
"test": "git submodule update --init --recursive; node test/run-tests.js",
```
This is a basic inital solution, it can be improved with followups later

takikawa:
> Once the tests have been moved to the main repo we can updated and land this patch. Do you know the timeline for this?

Thanks for the comments! We've now imported the tests into a more official repo (https://github.com/tc39/source-map-tests). I'll work on updating this PR to address your comments and update for the new location soon (hopefully this week).

takikawa:
I've update the PR and pointed it at the new repo, and addressed your comments (except one, commented above). Happy to address any further feedback.

test/test-spec-tests.js
```diff
@@ -0,0 +1,208 @@
+/* -*- Mode: js; js-indent-level: 2; -*- */
+/*
+ * Copyright 2024 Mozilla Foundation and contributors
+ * Licensed under the New BSD license. See LICENSE or:
+ * http://opensource.org/licenses/BSD-3-Clause
+ */
+
+const fs = require("node:fs/promises");
```
bomsy, line 8:
```suggestion
const fs = require("fs");
```

bomsy: This is likely causing the failure in the jobs. Can you fix.

bomsy, line 8:
```suggestion
const fs = require("fs").promises;
```

bomsy: So we try this to fix the failing job?

bomsy:
@takikawa  The patch  LGTM. Added a comment to resolve the failure in the jobs.

bomsy (approved):
LGTM! Lets land this Thanks @takikawa for the patch.

bomsy:
@takikawa Applying the fix to see if the 

> @takikawa The patch LGTM. Added a comment to resolve the failure in the jobs.

Fixed the failure in the jobs, so we can land the patch.

Issue #506, opened 2024-05-07
Original position lookup in an indexed source map can fail due to failure to find the right section


takikawa:
Here's an example test case:

```js
// add to test/test-source-map-consumer

exports["test a mapping at the boundary of indexed source map offset"] =
  async function (assert) {
    const map = await new SourceMapConsumer(util.indexedTestMapAtOffsetBoundary);
    util.assertMapping(1, 0, "/the/root/one.js", 1, 0, null, null, map, assert);
    map.destroy();
  };
```

```js
// add to test/util.js 

// This mapping is for testing a case where the mapped position is at the
// section offset.
exports.indexedTestMapAtOffsetBoundary = {
  version: 3,
  file: "min.js",
  sections: [
    {
      offset: {
        line: 0,
        column: 0,
      },
      map: {
        version: 3,
        sources: ["one.js"],
        sourcesContent: [
          "ONE.foo = function (bar) {\n   return baz(bar);\n };",
        ],
        names: ["bar", "baz"],
        mappings: "AAAA",
        file: "min.js",
        sourceRoot: "/the/root",
      },
    },
  ],
};
```

Currently it would fail like this:

```
FAILED ./test-source-map-consumer: test a mapping at the boundary of indexed source map offset!
AssertionError [ERR_ASSERTION]: Incorrect line, expected 1, got null
    at Object.assertMapping (/home/asumu/source-map/test/util.js:322:12)
    at exports.test a mapping at the boundary of indexed source map offset (/home/asumu/source-map/test/test-source-map-consumer.js:2190:10)
    at async run (/home/asumu/source-map/test/run-tests.js:21:11)
```

In this test, the index map has a section at offset line 0, column 0. The section has a single mapping from 0, 0 -> 0, 0. This should behave the same as a regular source map with just that one section.

However, the test actually fails because the section lookup fails. I think this is actually due to an off-by-one error in the section lookup code, which I'll submit a PR for.

PR #507, opened 2024-05-07
Fix an off-by-1 error in section lookup

takikawa:
This should fix issue #506.

### DIFF
--- a/lib/source-map-consumer.js
+++ b/lib/source-map-consumer.js
@@ -842,8 +842,11 @@ class IndexedSourceMapConsumer extends SourceMapConsumer {
           return cmp;
         }
 
+        // The generated column is 0-based, but the section offset column is
+        // stored 1-based.
         return (
-          aNeedle.generatedColumn - section.generatedOffset.generatedColumn
+          aNeedle.generatedColumn -
+          (section.generatedOffset.generatedColumn - 1)
         );
       }
     );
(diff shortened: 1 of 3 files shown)


lib/source-map-consumer.js
```diff
@@ -842,8 +842,10 @@ class IndexedSourceMapConsumer extends SourceMapConsumer {
           return cmp;
         }
 
+        // The needle column is 0-based, but the section offset column is
```
bomsy, line 845:
```suggestion
        // The generated column is 0-based, but the section offset column is
```

bomsy: I'm thinking that saying generated column feels more accurate here.

lib/source-map-consumer.js
```diff
@@ -842,8 +842,10 @@ class IndexedSourceMapConsumer extends SourceMapConsumer {
...
+        // stored 1-based.
         return (
-          aNeedle.generatedColumn - section.generatedOffset.generatedColumn
+          aNeedle.generatedColumn + 1 - section.generatedOffset.generatedColumn
```
bomsy, line 848:
```suggestion
          aNeedle.generatedColumn - (section.generatedOffset.generatedColumn - 1)
```

bomsy: Since according the the spec, the columns are mostly 0-based and in this case it is the section offset column was stored as 1-based 
for VLQ encoding compatibility. Lets do the adjustment on the section offset column rather than on the generated column. WDYT?

bomsy (changes requested):
Thanks for this patch. Apologies for the delay in reviews. It  looks great! I've got a few review comments.

takikawa:
> Thanks for this patch. Apologies for the delay in reviews. It looks great! I've got a few review comments.

Thanks for the suggestions! I've updated the patch now.

bomsy (approved):
Thanks for the updates. LGTM!

PR #517, opened 2025-07-08
Remove whatwg-url module

### LINKED ISSUE #400: Consider removing whatwg-url polyfill

nchevobbe:
The native URL constructor does handle custom schemes just fine now, we don't need to have this module anymore.

Fixes #400.


bomsy (approved):
Lgtm!

PR #519, opened 2025-07-30
Update CHANGELOG.md

nchevobbe:
Add information for 0.7.4 , 0.7.5 and 0.7.6


lukaselmer:
@nchevobbe is it possible to release a newer version (`0.8.0`?), so that `pnpm up source-map` does the right thing? It currently updates to the beta branch, which is probably not what people want 🤔 

image

PR #532, opened 2026-09-08
Allow initializing WASM contents in Node.js

### LINKED ISSUE #527: allow wasm to be bundled
# Support bundlers (esbuild, webpack) without requiring internal module swaps for WASM loading

## Problem

When bundling `source-map` with esbuild (platform: node), the WASM loading breaks in production because:

1. `read-wasm.js` uses `path.join(__dirname, "mappings.wasm")` — but after bundling, `__dirname` points to the output directory, not `node_modules/source-map/lib/`
2. `initialize()` is a no-op in the Node version (`read-wasm.js`), so there's no way to provide the WASM from the consumer side without swapping to `read-wasm-browser.js` via a bundler resolve plugin
3. The `"browser"` field in `package.json` maps `read-wasm.js` → `read-wasm-browser.js`, but esbuild with `platform: node` (correctly) ignores browser fields

This forces users to either:
- Copy `mappings.wasm` to their output directory with the exact filename (fragile, undocumented contract)
- Use a resolve plugin to swap `read-wasm.js` → `read-wasm-browser.js` and manually call `initialize()` (depends on internal file paths)

## Proposal

Make `initialize()` work regardless of platform — if the user provides WASM via `initialize()`, use it; otherwise fall back to the `fs.readFile(__dirname)` approach. This is essentially making the Node `read-wasm.js` behave like:

```js
let initialized = null;

module.exports = function readWasm() {
  if (initialized) return Promise.resolve(initialized);
  // existing fs.readFile fallback
  return new Promise((resolve, reject) => {
    const wasmPath = path.join(__dirname, "mappings.wasm");
    fs.readFile(wasmPath, null, (error, data) => {
      if (error) reject(error);
      else resolve(data.buffer);
    });
  });
};

module.exports.initialize = function (input) {
  initialized = input;
};
```

This way bundler users can call `SourceMapConsumer.initialize({"lib/mappings.wasm": wasmBuffer})` on Node without needing to swap modules.

## Alternatives considered

- Exporting `mappings.wasm` via package.json `"exports"` so bundlers can resolve it cleanly
- Shipping a pre-built JS fallback (like `source-map-js` does — no WASM at all)

## Environment

- source-map 0.7.6
- esbuild 0.28.x, platform: node, format: esm
- Node.js 24.x
- Deploying to Docker (only `dist/` is copied — no `node_modules`)

Hanabi9248:
Fixes #527.

`SourceMapConsumer.initialize()` ignores supplied WASM bytes in Node.js, so a bundle still tries to read `mappings.wasm` next to the relocated module.

Use a supplied `ArrayBuffer` when available, keeping the filesystem loader as the default. The README and API comments now describe this Node.js behavior. Callers must initialize before constructing consumers; an existing WASM instance is not replaced.

The regression test starts a fresh process, makes filesystem loading fail, and checks a source position using the real WASM. It fails before this change and passes afterward. Separate tests cover the supplied bytes and default loader.

Tested on Windows with Node.js v24.19.0:

- `node test/run-tests.js`: 193/193 passed.
- ESLint on the changed JavaScript files: passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,12 @@ sourceMap.SourceMapConsumer.initialize({
 });
 ```
 
+In node.js, initialization is optional: the module loads its bundled
+`lib/mappings.wasm` file by default. When a bundler relocates the module, you can
+instead provide an `ArrayBuffer` containing the WASM bytes. Call `initialize`
+before constructing any `SourceMapConsumer`s. URL initialization is only supported
+outside of node.js.
+
 #### new SourceMapConsumer(rawSourceMap)
 
 The only parameter is the raw source map (either as a string which can be
--- a/lib/read-wasm.js
+++ b/lib/read-wasm.js
@@ -6,7 +6,13 @@
 const fs = require("fs");
 const path = require("path");
 
+let mappingsWasm = null;
+
 module.exports = function readWasm() {
+  if (mappingsWasm instanceof ArrayBuffer) {
+    return Promise.resolve(mappingsWasm);
+  }
+
   return new Promise((resolve, reject) => {
     const wasmPath = path.join(__dirname, "mappings.wasm");
     fs.readFile(wasmPath, null, (error, data) => {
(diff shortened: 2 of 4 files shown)


Hanabi9248:
Following up on the Node.js bundling case in #527: would supporting explicitly supplied WASM bytes through initialize() fit the API direction? This keeps filesystem loading as the default and tests the bundled case in a fresh process with filesystem loading disabled. I’d appreciate feedback when you have time.

PR #536, opened 2026-09-09
Fix wasm-mappings build failures and warnings

bomsy:
This patch
 - Fixes wasm32 callback import linkage in source-map-mappings-wasm-api
 - Sets the rust edition for the package to 2024 (latest) and fixes related warnings accordingly
 - Tries to cleanup most of the build warnings


bomsy:
Closing this, as we a re taking a different approach.

